PR #66, opened 2026-04-26
User-selectable English variant for device TTS

mikelward:
## Summary

Adds a `VoiceLocale` preference (System / en-US / en-GB / en-AU) shown in Settings under the device TTS engine. Default is `System`, which preserves current behaviour (`Locale.getDefault()` → best matching voice). Picker only surfaces for the `DEVICE` engine — OpenAI/Gemini voices have fixed accents, so a regional control there would be misleading. Routed through `FetchAndNotifyWorker.speakWithFallback` and the Settings preview path so a tap immediately plays back in the chosen accent.

The OpenAI default voice is now also locale-aware:

- `EN_GB` → `fable` (the only British voice in OpenAI's lineup)
- `EN_US` / `EN_AU` / `SYSTEM`-on-non-en-GB → `nova`

This is a *default-only* change — once a user explicitly picks an OpenAI voice in Settings, that choice persists regardless of locale changes. Existing users who never picked a voice silently move from `alloy` to the new locale-appropriate default.

OpenAI has no Australian-accented voice, so en-AU falls back to `nova` rather than to British `fable`. Gemini's voices are language-agnostic personalities and aren't remapped per locale.

## Known follow-ups (not blocking this PR)

- `SettingsState`'s initial `openAiVoice` is `UserPreferences.DEFAULT_OPENAI_VOICE` (`alloy`), so the picker may briefly render `alloy` for the few hundred ms before the first DataStore emission overrides it. Cosmetic only — the worker always reads from `prefs` so morning playback uses the right voice.

## Test plan

- [ ] Open Settings, pick Device engine — picker appears with four options
- [ ] Switch picker to en-GB on a US-locale phone, tap Test voice — UK accent plays
- [ ] Switch picker to en-AU, tap Test voice — AU accent plays (where available on the device)
- [ ] Set picker to System, change phone language to en-GB, fire morning insight — speaks with UK accent
- [ ] Switch to OpenAI engine on en-US — voice picker pre-selects `nova`
- [ ] Switch picker to en-GB while on OpenAI — voice picker pre-selects `fable`
- [ ] Pick a different OpenAI voice explicitly, switch picker to en-GB — explicit choice persists (not overridden)
- [ ] Existing users without the new key in DataStore default to `SYSTEM` (no migration required)

https://claude.ai/code/session_01T5P8n8dTywRkQqFfTCBUWS


mikelward:
Closing: the VoiceLocale / English-variant feature landed on main across many subsequent commits. This branch is 538 commits behind and fully superseded.

PR #115, opened 2026-04-27
:core — clarify slicedForToday / slicedForTonight empty-slice docs

mikelward:
## Why

Follow-up to PR #112. Copilot's post-merge review caught a KDoc-vs-behavior
mismatch on `slicedForToday`:

> KDoc says it "falls back to the original DailyForecast when the slice would
> be empty", but the implementation returns `copy(hourly = emptyList())` when
> the filtered slice is empty.

Correct diagnosis. The same wording slip exists on `slicedForTonight` (which
was the model I copied for the docs).

Copilot suggested two fixes — (a) actually return `this`, or (b) update the
KDoc. **(a) would defeat the slicing**: returning the original keeps the
out-of-window hourly entries, which would re-surface them to
`RenderInsightSummary`'s `peakPrecip()` rule — exactly the bug PR #112 was
written to fix. So (b) it is.

## What

Doc-only on both `slicedForToday` and `slicedForTonight`: describe the actual
behavior (keep day-level aggregates, drop out-of-window hourly, fall through
to the noon / day-level fallback) and explicitly call out *why* `return this`
would be wrong, so the next reader doesn't repeat Copilot's hypothesis.

No behavior change. No tests changed.

## Test plan
- [ ] CI green (no behavior under test was touched).

https://claude.ai/code/session_01CBdNYt2ZPFf9FAC4CjZfuv

---
_Generated by [Claude Code](https://claude.ai/code/session_01CBdNYt2ZPFf9FAC4CjZfuv)_


mikelward:
Closing as stale — predates the CI hardening in #488 and hasn't been touched in 14+ days. Reopen and rebase onto current `main` if you want to pick this back up.

---

PR #273, opened 2026-05-01
:app — hash crash file for ack identity, not mtime

mikelward:
## Summary

- Crash banner identity was the crash file's `lastModified()` mtime. Filesystem mtime resolution can be coarser than ms (1s on ext4), so two distinct crash logs in a fast crash / restart loop can share an mtime — after the first is acked, the second is silently swallowed and the banner never reappears.
- Switch to a SHA-256 content hash. The crash file already includes a ms-precision header line and the recent log buffer, so two real crashes hash differently even when their mtimes collide.
- Added a regression test (`mtime collision with new content still re-surfaces`) that pins the new mtime to match the old one and asserts the new crash still surfaces. Updated the existing "fresh crash" test to drop the now-unnecessary `setLastModified` nudge, and added an "empty ack file falls back to surfacing" guard for truncated writes.

## Test plan

- [ ] CI: `JVM unit tests` green (covers `:app:testDebugUnitTest` including `CrashAckTest`).
- [ ] Manual: trigger two crashes back-to-back on device, ack the first, verify the second's banner appears on the next launch.

https://claude.ai/code/session_011bhAmNdwZdpXhZuveQn4hn

---
_Generated by [Claude Code](https://claude.ai/code/session_011bhAmNdwZdpXhZuveQn4hn)_

### DIFF
--- a/app/src/main/kotlin/app/clothescast/diag/DiagLog.kt
+++ b/app/src/main/kotlin/app/clothescast/diag/DiagLog.kt
@@ -5,6 +5,7 @@ import android.util.Log
 import java.io.File
 import java.io.PrintWriter
 import java.io.StringWriter
+import java.security.MessageDigest
 import java.text.SimpleDateFormat
 import java.util.ArrayDeque
 import java.util.Date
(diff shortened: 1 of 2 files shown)


mikelward:
Closing as stale — predates the CI hardening in #488 and hasn't been touched in 14+ days. Reopen and rebase onto current `main` if you want to pick this back up.

---

PR #301, opened 2026-05-03
:app + :core — clothes-rule thresholds honour user's temperature unit

mikelward:
## Why

The Settings → Clothes rules editor was hardcoded to °C: rows showed `%.0f°C` regardless of preference, the edit-dialog field was labelled `Threshold (°C)`, and the typed value was stored as Celsius even for °F users. The Today screen's rationale dialog already converted (`TodayScreen.kt:642–693`) — this PR closes the parity gap.

## Approach

Per the chosen design, each `ClothesRule.TemperatureBelow` / `TemperatureAbove` now carries the unit it was *entered in*, so a `65°F` threshold stays exactly `65°F` across °C↔°F switches (no integer round-trip drift). The `matches()` comparison converts to Celsius at runtime; the DTO's new `unit` field is nullable so pre-upgrade installs deserialise as `CELSIUS` and behave exactly as before.

UI:
- `ClothesContent` is plumbed `temperatureUnit` from `SettingsState`.
- `describeCondition` renders the **current** unit first; if the rule's stored unit differs, the original is appended in parentheses — e.g. a 65°F rule viewed under °C reads `18°C (65°F)`.
- The edit dialog pre-fills in the current display unit and saves the new value tagged with that unit. Field label uses `unit.symbol()`.

i18n: `settings_clothes_value_label_temp_c` → `settings_clothes_value_label_temp` with a `(%1$s)` symbol placeholder; the two `settings_clothes_cond_temp_*` strings drop hardcoded `%.0f°C` for `%1$s`. Kotlin pre-formats the value+symbol; all 47 locales now render unit-agnostically.

## Tests

- `:core:domain` — new `UnitsTest.kt` covering `toUnit`/`fromUnit` round-trips; `ClothesRuleTest` gains Fahrenheit-typed below/above cases plus a default-unit guard.
- `:app` — `SettingsRepositoryTest` covers a Fahrenheit-rule round-trip and a legacy-JSON-without-`unit` deserialisation case.
- New `SettingsClothesFahrenheitPreview` Roborazzi snapshot exercises the mixed-unit row rendering.

## Privacy

No change to what crosses the device boundary — this is purely on-device threshold storage and rendering.

## Test plan

- [ ] CI's `JVM unit tests` job (incl. domain + DTO + Roborazzi snapshot) is green.
- [ ] CI's `Android debug build` job confirms Compose code compiles.
- [ ] Manual: switch Region → °F. Settings → Clothes shows `when min temperature is below 18°C (65°F)` for the legacy default sweater rule. Edit, save 60 — row becomes `when min temperature is below 60°F` (no parens). Switch back to °C — row reads `when min temperature is below 16°C (60°F)`.

https://claude.ai/code/session_01JYEviNdLQpAVv44dFnqfut

---
_Generated by [Claude Code](https://claude.ai/code/session_01JYEviNdLQpAVv44dFnqfut)_


mikelward:
Codex's P1 was a real bug — fixed in `e5d1017`. The dialog now guards against the no-op confirm path: when the typed integer matches the rounded pre-fill *and* the condition type is unchanged, it returns the original `condition` verbatim (preserving its stored unit), instead of re-tagging with the current `temperatureUnit`. So a 65°F rule opened under °C, OK'd without changes, stays `TemperatureBelow(65.0, FAHRENHEIT)` rather than drifting to `18°C → 64°F`.

---

PR #317, opened 2026-05-04
:app + :core — outfit picker reads ClothesRule directly, drop the parallel Thresholds knob

mikelward:
## Summary

The "Why this outfit?" dialog was reading numbers from a separate
`OutfitSuggestion.Thresholds` preference whose 22°C shorts default
disagreed with the user's clothes-rule shorts cutoff (24°C default).
The user reported `Trousers · Feels-like high only 15°C at 14:00,
under 22°C` even though their threshold is 24. Two threshold systems,
no UI to bridge them.

This PR removes `OutfitSuggestion.Thresholds` entirely and points
`OutfitSuggestion.fromForecast` / `explainFromForecast` at
`prefs.clothesRules`. The home-screen icon and the bulleted
recommendations now read the same numbers — one mental model.

## What changes

- **Top:** firing `jacket` / `coat` rule → `THICK_JACKET`; firing
  `sweater` / `hoodie` → `SWEATER`; otherwise `TSHIRT`. Mapping is a
  small priority list of catalog item keys, so a user who deleted
  (say) the sweater rule still gets the right icon when their jacket
  rule fires.
- **Bottom:** firing `shorts` rule → `SHORTS`, else `LONG_PANTS`. The
  old "min ≥ 15°C" floor (the *afternoon-only shorts* guard) is gone —
  single threshold, matches what the user typed.
- Operators match each rule's existing semantics: `TemperatureBelow`
  (strict `<`) for the cold tiers, `TemperatureAbove` (strict `>`) for
  shorts.
- `Fact.thresholdKind` enum → `Fact.ruleItem` string keyed to
  `ClothesRule.item`.
- `UserPreferences.outfitThresholds` dropped.
- `SettingsRepository.adjustOutfitThreshold` /
  `setOutfitThresholds` / `resetOutfitThresholds` + the four
  `OUTFIT_THRESHOLD_*` keys gone, replaced by a single
  `adjustClothesRuleThreshold(item, deltaC)` that read-modify-writes
  inside one `DataStore.edit` transaction (so tap-spam doesn't drop
  intermediate writes).
- `adjustClothesRuleThreshold` preserves the rule's stored unit (a
  75°F rule stays in °F on disk after a `+1°` tap), and re-creates
  the rule from `ClothesRule.DEFAULTS` if the user previously deleted
  it rather than no-op'ing.
- The "Reset thresholds" button + its translations are gone — Settings
  → Clothes is the canonical management surface; the dialog is now
  purely a quick-nudge over the deciding cutoff.
- `InsightCache` schema bumped (`latest_insight_v4` → `v5`,
  `latest_tonight_insight_v3` → `v4`); pre-bump payloads drop and the
  next worker run regenerates.

## Behaviour notes

With default rules (sweater 18, jacket 12, coat 6, shorts 24), the
home-screen icon now picks `THICK_JACKET` below 12°C (was below 8°C)
and `SHORTS` above 24°C with no min-floor (was `max ≥ 22 AND min ≥
15`). Two `GenerateDailyInsightTest` expectations updated to match.

Per the user, no migration needed — no production users yet.

## Privacy

No change to what crosses the device boundary. Thresholds remain
local-only and only influence which icons render and the prose in the
rationale dialog.

## Test plan

- [ ] CI: `JVM unit tests` green (`:core:domain:test`,
      `:core:data:test`, `:app:testDebugUnitTest`).
- [ ] Roborazzi snapshot regen lands and the new
      `OutfitRationaleDialog*` PNGs render `under 24°C` / `under 18°C`
      with the rule-driven values.
- [ ] Manual: open Today → Why? on a `LONG_PANTS` day, confirm prose
      cites the user's `shorts` rule threshold (default 24°C, not 22).
- [ ] Manual: tap `+1°` in the rationale dialog, then go to Settings →
      Clothes and confirm the matching rule's value moved by 1° in
      whatever unit the rule was stored in.
- [ ] Manual: delete the `shorts` rule in Settings, return to Today,
      open the dialog, tap `+1°` — confirm a new `shorts` rule gets
      re-created from the catalog default with the adjusted threshold.

Local validation gap: this sandbox lacks the Android SDK and Google
Maven access, so even `./gradlew :core:domain:test` fails at root
config time. Leaning on CI for the unit-test pass.

https://claude.ai/code/session_01D9f7BBx3af4UsPnfYJkkCH

---
_Generated by [Claude Code](https://claude.ai/code/session_01D9f7BBx3af4UsPnfYJkkCH)_


mikelward:
Snapshot regen at `f21b3f9` — three modified PNGs, embedded inline since the mobile app's Files tab can't render binary diffs.

### `outfit_rationale_dialog.png`

Expected from this PR — the bottom fact now cites the `shorts` rule's `feelsLikeMax` cutoff (24°C) instead of the old `feelsLikeMin` floor (15°C).

Before (`1c2fbc0`):

outfit_rationale_dialog before

After (`f21b3f9`):

outfit_rationale_dialog after

### `outfit_rationale_dialog_tuned.png`

Expected from this PR — the "Reset thresholds" button is gone (Settings → Clothes is the canonical management surface now), and the bottom fact again switched from min-floor to the shorts rule's max cutoff.

Before (`1c2fbc0`):

outfit_rationale_dialog_tuned before

After (`f21b3f9`):

outfit_rationale_dialog_tuned after

### `launcher_icon_dev.png`

Unrelated to this PR's changes — flagging it for awareness rather than as something I expected to touch. Possibly a flaky/non-deterministic capture (font hinting, vector-rasterisation rounding) on the dev-only "DEV" badge that was added in `9f866e2`. Eyeball it; if it's a regression, worth a separate follow-up.

Before (`1c2fbc0`):

launcher_icon_dev before

After (`f21b3f9`):

launcher_icon_dev after

---

PR #325, opened 2026-05-05
:app + :core — humorous Gemini TTS voice directives + dev-only Custom field

mikelward:
## Summary

- Adds 15 character / persona registers to `TtsStyle` (Pirate, Cowboy, Shakespearean, Surfer, Parent, Child, Teenager, Grandparent, Kids' TV host, Talkback host, Morning DJ, Science teacher, Historian, Sportscaster, Weather forecaster) on top of the existing NORMAL / NEWSREADER baseline. Each is a `"Read the following as <persona>…"` preamble that shapes Gemini's vocal delivery and permits brief in-character interjections (`Arrr`, `Howdy`, `Forsooth`).
- Each directive ends with the existing "no audio effects, background noise, or vinyl-style texture" trailer — important for personas like PIRATE that might otherwise tempt the model to layer in ambience.
- Adds a **development-only** `TtsStyle.CUSTOM` with a free-text directive field in Settings, so directive wording can be iterated at runtime and promoted to a baked-in entry once it sounds right. Marked `TODO(pre-release)` at every touchpoint (enum entry, `UserPreferences.customTtsStyleDirective`, `CUSTOM_TTS_STYLE_DIRECTIVE` repo key + setter, ViewModel setter, the `if (ttsStyle == CUSTOM)` UI block, the helper composable). A blank CUSTOM falls back to the `NORMAL` preamble so a half-typed directive doesn't break TTS.

## Privacy

These directives are TTS preambles only — they shape *delivery*, not the prose. The TTS request already carries the rendered insight prose plus a directive preamble, so this doesn't broaden what leaves the device.

## Notes

- The Style picker is Gemini-only (existing UI gating); on-device TTS doesn't accept style prompts.
- The picker dialog uses an existing scrollable column — 17 entries fit fine.
- Plan: `/root/.claude/plans/can-we-add-some-imperative-biscuit.md`.

## Test plan

- [ ] CI: `JVM unit tests` job — new tests in `GeminiTtsClientTest` verify each directive's signature phrase lands in the request body; CUSTOM round-trips a verbatim directive; blank CUSTOM falls back to NORMAL. The extended `tts style setter round-trips` iterates every enum value so future additions can't silently break the DataStore decode path.
- [ ] CI: `Android debug build` job — `:app:assembleDebug` covers the Compose `VoiceContent` signature change, the new `CustomDirectiveField`, and the `SettingsPreviews` updates.
- [ ] Local Gradle could not run in this sandbox (Google Maven is blocked, AGP doesn't resolve). Validation surface is CI.
- [ ] Manual on-device, once a build is available: Settings → Voice → Gemini engine → Style → pick Pirate → Test voice → confirm pirate-flavoured delivery on a few personas. Pick Custom, type a directive, Test, confirm. Pick Custom with the field blank, Test, confirm it still plays (NORMAL fallback).

https://claude.ai/code/session_01HfUMsyo3cfzDS7VvXhuHQy

---
_Generated by [Claude Code](https://claude.ai/code/session_01HfUMsyo3cfzDS7VvXhuHQy)_


mikelward:
Re Codex's P2 about `TtsStyle.CUSTOM` reaching release builds — keeping it broadly available for now on purpose. clothescast is BYOK for Gemini (the user supplies their own API key), so the cost / abuse surface for a freeform directive sits on the user's own quota, not ours, and the directive is stylistic rather than data-exfiltrating. Tester builds are release-style (not debug), and we want testers to be able to iterate on directive wording so good ones can be promoted to baked-in `TtsStyle` entries — a `BuildConfig.DEBUG` filter would lock testers out of exactly that loop.

The `TODO(pre-release)` markers (enum entry, `UserPreferences.customTtsStyleDirective`, repository key/setter, ViewModel setter, `if (ttsStyle == CUSTOM)` UI block, `CustomDirectiveField` composable) are the removal gate before public release. Will revisit if CUSTOM ends up sticking around longer than the tester window.

---

PR #327, opened 2026-05-05
:app + :core — Firebase Analytics + Crashlytics behind a Privacy toggle

mikelward:
## Summary

Wires up the Firebase Analytics + Crashlytics integration that PRIVACY.md
has been pre-disclosing since 2026-05-03. Default-on so the long tail of
installs report crashes without each user finding a toggle, with a
one-time non-blocking notice on the Today screen pointing at the new
Settings → Privacy sub-page where they can flip it off.

## Privacy callout (per CLAUDE.md)

This change opens a new device-boundary path. The hard limits in
PRIVACY.md's "Analytics and crash reporting" section are unchanged and
honoured in code:

- **Never sent:** calendar event data (titles / times / locations /
  attendees / "any events at all"), location coordinates or geocoded
  place names, insight prose, notification text, API keys, precise
  GPS, advertising IDs.
- **Sent (only what Firebase auto-collects):** stack traces, app /
  Android / device version, a non-resettable install ID for crash
  grouping; for Analytics, lifecycle events (`session_start`,
  `app_open`, `first_open` etc.) and screen views.

The new `Telemetry` controller deliberately does **not** call
`FirebaseCrashlytics.setCustomKey(...)` from any app data, and does
**not** push user-content into Analytics event params. Custom-event
instrumentation (per-setting analytics, etc.) is a follow-up PR; this
one is just plumbing + the toggle.

## How it's wired

- **`Telemetry`** (`app/diag/Telemetry.kt`) subscribes to
  `SettingsRepository.preferences.telemetryEnabled` and pushes each
  change into `FirebaseAnalytics.setAnalyticsCollectionEnabled` +
  `FirebaseCrashlytics.setCrashlyticsCollectionEnabled`. Both SDK
  flags persist across launches, so a flip-off survives process death.
  No-ops if Firebase didn't initialise (no `google-services.json`).
- **`TelemetryNoticeBanner`** on Today follows the existing
  `LocalBuildBanner` pattern — collected via DataStore, dismisses with
  an X (acks the notice) or "Open Privacy settings" (acks + deep-links).
- **`PrivacyContent`** is a new sub-page under a new
  `SettingsRoute.Privacy`, with the toggle, an inline summary of the
  hard limits, and a link to PRIVACY.md.
- **`telemetryEnabled` (default true)** + **`telemetryNoticeAcked`
  (default false)** persist in the existing settings DataStore.

## Build wiring

The `gms-google-services` + `firebase-crashlytics` Gradle plugins are
[...]

PR #328, opened 2026-05-05
:ci — decode google-services.json from secret so CI builds carry Firebase config

mikelward:
## Summary

Follow-up to #327. Materialises `app/google-services.json` on the CI
runner before Gradle runs, so the Firebase Gradle plugins
(conditionally applied per `app/build.gradle.kts`) actually take
effect.

Without this, CI APKs distributed via FAD had Firebase code linked but
no config resources, so `Telemetry` no-op'd at runtime via
`FirebaseApp.getApps(...).isEmpty()`. With it, Crashlytics + Analytics
report from CI debug APKs (and from the eventual release AAB) gated on
the per-user toggle in Settings → Privacy.

## Required GitHub secret

- **`GOOGLE_SERVICES_JSON`** — paste the contents of
  `app/google-services.json` directly into the secret value. (Base64
  is also accepted; the step auto-detects which shape was pasted.)
  Must include `client[]` entries for both `app.clothescast` (release)
  and `app.clothescast.debug` (debug variant) so both build types pass
  the gms plugin's "no matching client" check.

## Pattern

Mirrors the existing `DEBUG_KEYSTORE_BASE64` / `UPLOAD_KEYSTORE_BASE64`
decode steps but drops the `_BASE64` suffix because
`google-services.json` is text, not binary — pasting it raw works and
is one fewer step at setup time. Auto-detect inspects the first
non-whitespace character: a leading `{` is treated as raw JSON; anything
else is base64-decoded.

- Quiet skip when the secret isn't set (fork PR / fresh repo) — build
  still completes, telemetry just stays off.
- Up-front sanity check (`project_info` + `client` keys present) so a
  garbage / truncated paste fails here with a clear message, not as an
  opaque gms-plugin "JSON parse error" several minutes into the gradle
  run.

## Privacy

The data-boundary contract in PRIVACY.md is unchanged. This change just
flips Firebase from "linked, runtime no-op" to "linked, runtime active"
on CI builds — the runtime toggle in Settings → Privacy still controls
whether anything actually leaves the device.

## Test plan

- [ ] CI: both jobs green on this PR (the auto-detect handles the
  user's raw-JSON-paste secret).
- [ ] After merge: confirm `processDebugGoogleServices` task runs in
  the Android debug build job's gradle log.
- [ ] After merge: install the FAD-distributed debug APK, force a
  crash, check Firebase Crashlytics console for the report (with
  telemetry on in the new Privacy settings).
- [ ] After merge: flip telemetry off in Settings → Privacy, force
  another crash, confirm it does NOT appear in Crashlytics.

## Open PRs in the stack

- https://github.com/mikelward/clothescast/pull/328 — this PR

🤖 https://claude.ai/code/session_01NPW4Sv61Y65RzPB2pHzDb5

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,56 @@ jobs:
         with:
           cache-read-only: ${{ github.ref != 'refs/heads/main' }}
 
+      - name: Decode google-services.json from secret
+        # The Firebase Gradle plugins (gms-google-services + crashlytics) are
+        # applied conditionally on `app/google-services.json` being present —
+        # see app/build.gradle.kts. Materialising the file here makes the
+        # plugins apply on CI, so unit-tests and android-build see the same
+        # plugin state. Without this, the file is absent in CI sandboxes
+        # (it's .gitignore-d) and the SDKs no-op via
+        # FirebaseApp.getApps(...).isEmpty() at runtime — which is correct,
+        # just silently telemetry-less. Skipped quietly when the secret
+        # isn't set (fork PR / fresh repo).
+        #
+        # Auto-detects whether the secret holds raw JSON or base64. This
+        # repo's other binary secrets (DEBUG_KEYSTORE_BASE64,
+        # UPLOAD_KEYSTORE_BASE64) must be base64 because they're binary;
+        # google-services.json is text, so pasting it raw works too and is
+        # one fewer step at setup time. Either shape is accepted.
+        env:
+          GOOGLE_SERVICES_JSON: ${{ secrets.GOOGLE_SERVICES_JSON }}
+        run: |
+          if [ -z "$GOOGLE_SERVICES_JSON" ]; then
+            echo "No GOOGLE_SERVICES_JSON secret available; build will skip Firebase plugins."
+            exit 0
+          fi
+          # Strip surrounding whitespace so a stray newline at the end of a
+          # paste (or default 76-col-wrapped base64) doesn't trip strict
+          # decoders. A leading `{` after stripping means the user pasted
+          # raw JSON — write it through. Otherwise treat as base64.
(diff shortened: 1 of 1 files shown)


mikelward:
Both jobs failed at the new "Decode google-services.json from secret" step with exit code 1. The original validation grep tripped, which means the decoded file didn't contain `"project_info"` / `"client"` — most likely the secret value was either pasted as raw JSON (not base64) or had wrapping whitespace that confused `base64 --decode`.

Pushed `493ba87` to handle both shapes:

- **Auto-detect**: if the secret value's first non-whitespace char is `{`, treat as raw JSON and write through. Otherwise base64-decode. Either way, you don't have to re-create the secret unless you want to.
- **Whitespace strip** before decode (handles wrapped-paste base64).
- **Distinct errors** for "decode failed" vs "wrong JSON shape", with byte counts and the exact one-liner to regenerate the secret.

If `GOOGLE_SERVICES_JSON_BASE64` was added under a different name in repo secrets, the step will quietly skip — confirm the name matches if you still see "build will skip Firebase plugins" in the new run.

---

mikelward:
Force-pushed `ba476fd` dropping the `_BASE64` suffix from the env-var name (text file, not binary, so the suffix was misleading).

⚠️ **Action needed if your existing secret is named `GOOGLE_SERVICES_JSON_BASE64`** — GitHub doesn't support renaming secrets, so you'd need to:

1. Repo → Settings → Secrets and variables → Actions
2. Delete `GOOGLE_SERVICES_JSON_BASE64`
3. Re-add as `GOOGLE_SERVICES_JSON` with the same raw JSON value

If you already named it `GOOGLE_SERVICES_JSON`, you're set. The new CI run on `ba476fd` will tell — look for `Detected raw JSON in secret` (good) vs `No GOOGLE_SERVICES_JSON secret available` (rename needed) in the JVM unit tests log.

---

PR #338, opened 2026-05-05
:app + :core:domain — trim Gemini voice list to Despina, Leda, Iapetus, Charon

mikelward:
Trims `GEMINI_VOICES` from 11 entries down to the four requested: Iapetus, Charon, Despina, Leda.

Also fixes two housekeeping issues uncovered along the way:
- `SettingsRepositoryTest` was asserting `geminiVoiceDefault == "Erinome"` — stale since the default was set to `"Leda"` in `Settings.kt` and never updated in the test.
- Two in-code doc comments cited `"Erinome"` / `"Kore"` as example voice names; updated to `"Leda"` / `"Iapetus"`.

No behaviour change for users who haven't explicitly picked a voice — default remains `"Leda"`. Users who had one of the removed voices stored in DataStore will fall back to `"Leda"` on next load (existing fallback logic in `SettingsRepository`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01LKEo19ntZq2qp6rDw5jinB)_


mikelward:
Closing: the voice-list trim landed on main via a different approach (7 voices, 33ecd10) that superseded this 4-voice cut.

PR #377, opened 2026-05-12
Surface a louder callout when the forecast models disagree

mikelward:
## Summary

When the cross-model confidence drops to `LOW` (today's apparent-max temperature or peak rain chance vary widely across ECMWF / GFS / ICON), the existing `ConfidenceChip` inside the insight card is easy to miss — it's a compact always-visible status indicator, not a "you should second-guess this" prompt.

This PR adds a full-width `LowConfidenceCallout` above the outfit preview that only renders at `LOW`, surfacing the actual spread numbers and a "dress for the worse case" nudge. HIGH and MEDIUM are unchanged — the chip alone is the right loudness for those.

The callout uses the same `errorContainer` tint as the LOW chip so the two reinforce each other instead of fighting for attention.

## Why this and not the chart overlays?

The original plan for PR 2 was "callout + per-model chart overlays + new setting." Splitting that into 2a (this PR) and 2b (the chart overlays) per CLAUDE.md's "one concern per PR" — this one renders existing `ConfidenceInfo` that's already in the bundle (zero new data), and ships value immediately; PR 2b adds the per-model hourly fetch + setting + chart rendering.

This also acts as a follow-on validation for #376's diagnostic: if `LOW` confidence is ever computed but we still never see the callout, the bug-report log will tell us why.

## Privacy

No new data leaves the device. The callout consumes `ConfidenceInfo` that was already in the bundle and was already shown (less prominently) on the screen.

## Test plan

- [ ] CI green on `JVM unit tests` (new Roborazzi snapshot `low_confidence_callout.png`)
- [ ] CI green on `Android debug build`
- [ ] Manual: force a `LOW` reading (or just wait for one to occur in the wild) and verify the callout appears above the outfit row.

Stacked on #376 (merged). PR 2b will land separately with the chart overlays.

https://claude.ai/code/session_01RiLFGDXGtXB2jGqrmwRiPE

---
_Generated by [Claude Code](https://claude.ai/code/session_01RiLFGDXGtXB2jGqrmwRiPE)_


mikelward:
New snapshot from this push (CI regen `1821f99`):

`low_confidence_callout.png` (added):

low_confidence_callout

No other snapshots shifted — the new callout only renders at `ForecastConfidence.LOW`, so existing Today-screen snapshots (HIGH / MEDIUM / no-insight scenes) are byte-identical.

---

PR #385, opened 2026-05-13
Show per-model wind speed alongside the temperature chart

mikelward:
## Summary

Follow-up to #384. When models disagree on feels-like temperature, wind chill is the usual culprit below ~20 °C, and you can't see that on the temp chart alone.

- New `WindCard` under the existing Today temp + rain pair (gated on the existing "Show model spread" power-user setting).
- No "main" line — `HourlyForecast` doesn't carry wind — so the chart is purely per-model overlays in their pinned ECMWF / GFS / ICON colours. Card auto-hides when no consulted model returned wind data.
- Plumbs three new per-model hourly fields on the same batched Open-Meteo request (free; same network round-trip):
  - `wind_speed_10m` → `PerModelHour.windSpeedKmh`
  - `relative_humidity_2m` → `PerModelHour.relativeHumidityPct`
  - `cloud_cover` → `PerModelHour.cloudCoverPct`

Only the wind chart is wired up. Humidity / cloud are plumbed but not yet visualised — see TODOs.

## TODOs queued in `PerModelHourly.kt`

- **model-divergence-summary**: tappable "biggest contributing factor" hint per hour when the spread is wide.
- **model-spread-stat**: RMSE / std-dev alongside the current max-min spread for confidence.
- **main-line-model**: `OpenMeteoClient` doesn't pass `models=`, so the blended main line tracks whichever model `best_match` picks rather than a midpoint — confusing when models diverge. (Confirmed by inspection — `OpenMeteoClient.kt:90`.)

## Cache backwards-compat

The three new fields are nullable in the domain (not just the DTO), so legacy cached payloads keep the temp / rain overlay and only hide the wind chart until the next worker refresh repopulates.

## Privacy

No change to what leaves the device — same Open-Meteo lat/lon call, just three extra hourly fields requested on the existing round-trip.

## Test plan

- [ ] CI: domain + data + app JVM tests
- [ ] CI: snapshot regen for the new `wind_card_with_model_spread.png`
- [ ] On device with model spread on: wind card appears below rain card, three coloured lines diverge visibly across the day

Open PRs in this stack:
- https://github.com/mikelward/clothescast/pull/385 (this PR)

---
_Generated by [Claude Code](https://claude.ai/code/session_01DUHtB5jzSzfbnj3GGqyggP)_


mikelward:
New snapshot from `f0e3cbe`: `wind_card_with_model_spread.png` (only added file in the regen; nothing else moved bytes).

wind_card_with_model_spread @ f0e3cbe

---

PR #387, opened 2026-05-13
Filter the wind legend to the models that are actually plotted

mikelward:
## Summary

Follow-up to a Codex review comment on #385: the wind card's legend was
sourced from the unfiltered `perModelHourly`, but `WindChart` drops any
model without a full per-hour `windSpeedKmh` series before plotting.
For an older cached payload (or a model run with missing diagnostic
fields) the legend could list GFS / ICON even though no corresponding
wind line was drawn — chart mislabeled.

- Pre-filter `perModelHourly` in `WindCard` to the same models
  `WindChart` will actually plot.
- Pass the filtered overlay to both the chart and `ModelSpreadLegend`
  so the legend lists exactly the lines drawn.
- Card-visibility check switches to "did any model survive the filter?"
  so we no longer render an empty card when models have scattered wind
  nulls but none have a complete series.

Addresses: https://github.com/mikelward/clothescast/pull/385#discussion_r3232881612

## Privacy

No change to what leaves the device — same Open-Meteo data, just a UI
filter on what was already fetched.

## Test plan

- [ ] CI: domain + data + app JVM tests
- [ ] CI: Android debug build
- [ ] CI: snapshot of `wind_card_with_model_spread.png` should be byte-identical (preview's sample data has wind on every model)

---
_Generated by [Claude Code](https://claude.ai/code/session_01CdGc2kxokM3rd2ybo4G2xB)_


mikelward:
Snapshot regen diff for ed00205 — only `precipitation_card.png` changed. My fix is scoped to `WindCard` and doesn't touch `PrecipitationCard` or its preview data (`PrecipitationCardPreview` uses just `SAMPLE_HOURLY_RAINY` with no model overlay), so this looks like flaky / non-deterministic rendering rather than a meaningful diff. Posting both versions inline since the GitHub mobile app shows "Binary files not rendered" for PNG diffs:

**`precipitation_card.png` — previous (7ffbfaa, same as main):**



**`precipitation_card.png` — new (ed00205):**



If they look visually identical to you, this is a sub-pixel / anti-aliasing wobble I'd suggest just accepting; if there's a visible change I'll dig deeper.

---

mikelward:
Closing — this fix already landed on `main` as part of the broader `PerModelDiagnosticCard` refactor (commits `87cbdd9` and the surrounding stack). On `main`:

- `ModelSpreadLegend` now takes `visibleModelIds: List<String>` directly instead of a `PerModelHourly`, with a KDoc that explicitly cites the pre-refactor bug ("silently listed models whose lines had been filtered out of a chart — e.g. wind diagnostic lines for models that didn't report wind") — see `app/src/main/kotlin/app/clothescast/ui/today/TodayScreen.kt:1306`.
- The wind / cloud / humidity diagnostic cards all flow through `PerModelDiagnosticCard`, which computes `visibleModels = MODEL_DRAW_ORDER.filter { it in seriesByModel }` after filtering out models with no values for the picked metric, and passes those exact IDs to the legend — `app/src/main/kotlin/app/clothescast/ui/today/PerModelDiagnosticCard.kt:79-98`.

Rebase produced a real conflict and the diff would only re-do work already on main, so closing without merging rather than landing a no-op.

---

PR #388, opened 2026-05-13
Label the main forecast line as Best match and fix sparse diagnostics

mikelward:
## Summary

Two adjacent fixes around the model-spread chart legend, folded together because they share the same legend abstraction.

### H6 — label "Best match"

The temperature and precipitation charts' blended "main" line is whatever Open-Meteo's `forecast` endpoint default selects — its **"best match"** auto-selection — which routinely tracks a different model than the consulted ECMWF / GFS / ICON overlays. Pre-existing diagnostic charts left this main line unlabelled, so when models diverged (your rain-chart screenshot showed this dramatically — the grey main line nowhere near GFS or ICON), the chart looked broken.

Adds a `MainLineLegend` (colour + label) parameter to `ModelSpreadLegend`. Temp + rain cards pass theme primary + "Best match"; diagnostic cards leave it null.

This doesn't change *which* model the main line uses (Open-Meteo doesn't expose which underlying model best_match resolved to), but it makes the disagreement legible rather than mysterious.

### Sparse-diagnostic + filter bugs

The wind / cloud / humidity cards each had two latent bugs:

1. **Sparse data dropped whole models.** The chart filtered out any model whose non-null hours didn't exactly match `times.size`, so a single missing hour dropped the whole model from the chart.
2. **Legend mislabelled.** The legend was derived from `perModelHourly.byModel` directly, so it could list a model whose chart line had been filtered out.

Refactor: the three near-identical diagnostic cards (WindCard / CloudCard / HumidityCard) collapse into one shared `PerModelDiagnosticCard` composable that:

- builds `(originalIndex, value)` pairs per model and plots those at their original x-positions via Vico's explicit-x `series` overload — sparse hours become small bridge segments rather than dropping the whole line;
- passes the resulting `visibleModels` list to both the chart line provider *and* the legend, so they can't disagree by construction;
- auto-hides the whole card when every consulted model is missing the metric.

Net diff: **−300 lines** after merging three duplicated chart files (`WindChart` / `CloudChart` / `HumidityChart`) into one shared composable.

## Test plan

- [ ] CI: JVM + Android debug build green
- [ ] CI: snapshot regen for the temp + rain cards (now carry the extra "Best match" legend chip) and the three diagnostic cards (same visuals on full data; the abstraction is verified by snapshot equality)
- [ ] On device with model spread on: legend reads "Models: ● ECMWF ● GFS ● ICON · ● Best match"

## What's left from the original "model divergence" thread

- **model-divergence-summary** — per-hour "biggest contributing factor" hint when spread is wide. Actionable now that wind / cloud / humidity are plumbed.
- **model-spread-stat** — RMSE / std-dev alongside max-min for confidence.

Both tracked as TODOs in `core/domain/src/main/kotlin/app/clothescast/core/domain/model/PerModelHourly.kt`.

## Privacy

No change — same Open-Meteo lat/lon, same fields.

Open PRs in this stack:
- https://github.com/mikelward/clothescast/pull/388 (this PR)


mikelward:
Snapshot regen from `0629258`:

**Modified — `precipitation_card_with_model_spread`** (now carries the "Best match" legend chip alongside the consulted models)

Before (`0581373`, parent of this regen):

precipitation_card_with_model_spread @ 0581373

After (`0629258`):

precipitation_card_with_model_spread @ 0629258

**Added — `cloud_card_with_model_spread`** (preview was wired in #386 but the snapshot regen landed after the merge, so this is the first captured frame)

cloud_card_with_model_spread @ 0629258

**Added — `humidity_card_with_model_spread`** (same story as cloud)

humidity_card_with_model_spread @ 0629258

The temperature card's "Best match" legend chip isn't currently snapshot-tested — only `ForecastChartWithModelSpreadPreview` exists (charts-only) and not a wrapping `ForecastCardWithModelSpreadPreview`. Pre-existing gap; flagging for a follow-up rather than rolling into this PR.

---

PR #390, opened 2026-05-13
Pin diagnostic chart axis when a model is missing edge hours

mikelward:
## Summary

The wind / cloud / humidity diagnostic cards plot per-model overlays at their original time indices so a single missing hour becomes a small bridge segment instead of dropping the whole model's line. That worked fine when the gaps were interior, but when *every* visible model was missing the metric at the start or end of the day — the upstream API occasionally returns null for the trailing hours of a run that's still warming up, or for the leading hours when a new run hasn't published yet — the explicit-x series only carried the surviving indices. With no main / full-length series and no x-range provider, Vico's `Zoom.Content` fit those remaining points to the full card width, so a line that only covered (say) the last four hours of the day visually stretched across the whole 0–23 axis and silently misaligned with the other hourly cards' x positions.

## What changed

- `PerModelDiagnosticChart` now overrides `getMinX` / `getMaxX` on its `CartesianLayerRangeProvider` to pin the chart's x-range to `0..times.lastIndex` (both `YAxis.Percent` and `YAxis.AutoZeroBased` branches). Sparse series now sit at their real x positions with the missing leading or trailing hours visibly empty — which is the correct affordance ("the model didn't predict this part of the day"), not a stretched line pretending coverage it doesn't have. Dense (no-gap) data renders identically because the data range and the pinned range coincide.
- Adds a `WindCard · sparse trailing only` preview + Roborazzi snapshot that zeroes out `windSpeedKmh` for hours 00–19 across every model, so the trailing-only edge case is locked in a snapshot and any future regression in the x anchor shows up as a snapshot diff.

## Test plan

- [ ] CI: `JVM unit tests` — including the new `wind_card_sparse_trailing` Roborazzi snapshot.
- [ ] CI: `Android debug build`.
- [ ] Eyeball the regenerated `wind_card_sparse_trailing.png` (sparse line on the right edge, leading hours empty) vs. the existing `wind_card_with_model_spread.png` (unchanged — the dense fixture already spans 0–23, so the pin matches the inferred range).
- [ ] Confirm `wind_card_with_model_spread`, `cloud_card_with_model_spread`, `humidity_card_with_model_spread` snapshots are byte-identical to pre-change (the pinned x-range should match Vico's inferred range when data is dense).

## Privacy

No change — purely a rendering tweak on data already on device.

---
_Generated by [Claude Code](https://claude.ai/code/session_01USc84fXQiq7hF7REkhSGnq)_


mikelward:
Snapshot regen at `81c1efc` only added the new sparse-trailing preview — no existing snapshots changed, which confirms the x-range pin is a no-op for dense fixtures (the inferred and pinned ranges coincide when data spans `0..times.lastIndex`).

**New: `wind_card_sparse_trailing.png`** — every model is missing wind for hours 00–19; only hours 20–23 carry values. The line should sit on the right edge of the chart with the leading hours visibly empty.

wind_card_sparse_trailing @ 81c1efc

---

PR #395, opened 2026-05-13
Pass best_match into the consensus blender, not around it

mikelward:
## Summary

Codex caught a real bug on #394's merged code:

`OpenMeteoClient.fetchForecast` passed `multi?.hourly` to `blendConsensusHourly` **before** the best_match overlay was injected into `byModel`. So the blender saw only the consulted models (ECMWF / GFS / ICON) and in regions / hours where only one of them reported, the consensus returned null and production fell back to raw best_match — exactly the failure mode the consensus blend was supposed to fix.

Test fixtures pre-built the post-injection `PerModelHourly` directly, so `ConsensusBlendTest` passed even though production silently behaved differently. The function's docs and the reviewer's intent both say best_match counts as a model in the mean; this restores that.

Fix is just reordering the two assignments so `perModelWithBestMatch` is built first, then handed to `blendConsensusHourly`. Comment above the reordering calls out the ordering requirement so the bug can't sneak back in via casual refactoring.

## Test plan

- [ ] CI: JVM + Android debug build green
- [ ] Existing `ConsensusBlendTest` continues to pass (was already exercising the post-injection shape; no fixture change needed)
- [ ] On device: same fix as #394's intent — divergence days where consulted models agreed should produce the consulted recommendation rather than tracking the Auto outlier

## No new test added

The bug was at the orchestration layer (which has integration-only tests against HTTP fixtures), not at the function level. A regression test would need either an inline per-model JSON fixture or extracting the orchestration into a pure helper for unit-testing — both bigger than this fix. I've added a comment at the ordering site instead; the next refactor that swaps the order will trigger the same Codex review (and a human reading the comment).

Credit to Codex for the catch.

Open PRs in this session:
- https://github.com/mikelward/clothescast/pull/395 (this PR — fingers crossed it gets #395)

---
_Generated by [Claude Code](https://claude.ai/code/session_01DUHtB5jzSzfbnj3GGqyggP)_


mikelward:
Snapshot regen from `b08dcfc`: one modified, no additions. This PR's diff is `OpenMeteoClient.kt`-only — none of the preview composables changed — so the byte difference here looks like minor chart-rendering nondeterminism (overlapping line anti-aliasing) rather than a behavioural visual shift. Posting both versions to confirm.

**`forecast_chart_with_model_spread`**

Before (`d852d24`, base of this regen):

forecast_chart_with_model_spread @ d852d24

After (`b08dcfc`):

forecast_chart_with_model_spread @ b08dcfc

---

PR #401, opened 2026-05-13
Re-take chart snapshots when only the card header renders

mikelward:
## Summary

PR #400's CI regen committed a chart-less `precipitation_card.png` as the new baseline — header text rendered, Vico chart body didn't (34890 → 11393 bytes). A manual revert on that PR restored the previous snapshot, but the existing retry path (f6fb4d6 *"Re-take UI snapshot captures automatically when they come out blank"*) would have happily landed it again on the next regen: `looksEmpty` only flags PNGs where every sampled pixel is identical, so a partial blank with rendered header text slips straight through.

## What

Test-infra only — no production code touched, no privacy-relevant change.

- New `captureChart(content)` helper that re-uses the existing retry/backoff plumbing but swaps in a stricter checker.
- New `chartLooksEmpty(file)` checker: subsumes `looksEmpty`, then samples only the bottom 60% of the image on a 24×16 grid and short-circuits as soon as three distinct colours are seen. Every real chart paints axis labels, grid lines and at least one data line down there — 3+ unique colours land easily — while a chart-less card body is one flat colour and trips the retry.
- The twelve chart-card snapshot tests route through `captureChart`:
  - `forecast_chart`, `forecast_chart_dark`, `forecast_chart_with_model_spread`, `forecast_card_with_model_spread`
  - `precipitation_card`, `precipitation_card_dark`, `precipitation_card_dry`, `precipitation_card_with_model_spread`
  - `wind_card_with_model_spread`, `wind_card_sparse_trailing`, `cloud_card_with_model_spread`, `humidity_card_with_model_spread`
- Non-chart captures keep the original `looksEmpty` so simple snapshots (launcher icons, settings screens, banners) aren't impacted.

## What this does *not* fix

The Vico race itself. When `CartesianChartModelProducer.runTransaction` doesn't deliver before `waitForIdle()` returns, we still retry up to three times — but now a partial blank no longer slips past as the committed baseline. A structural fix (pre-populating the producer eagerly so there's no async hop on first frame) was discussed and deferred — this lands the cheap defensive backstop first.

## Test plan

- [ ] CI: `JVM unit tests` green (the twelve chart-card tests should produce identical PNGs to their current baselines — no pixels change, only the empty-detection guard tightens).
- [ ] Manual: if a future regen produces a chart-less card, the retry's stderr warning surfaces in the CI logs instead of the bad PNG silently becoming the new baseline.

---
_Generated by [Claude Code](https://claude.ai/code/session_012jCU1oNPKJndBtLAbLAimX)_

### DIFF
--- a/app/src/test/kotlin/app/clothescast/ui/today/PreviewSnapshots.kt
+++ b/app/src/test/kotlin/app/clothescast/ui/today/PreviewSnapshots.kt
@@ -128,6 +128,22 @@ class PreviewSnapshots {
         }
     }
 
+    // Same as `capture`, but with a stricter blank check tuned for chart cards.
+    // The default `looksEmpty` only flags PNGs where every sampled pixel is
+    // identical, which misses the common chart race where the card header
+    // ("Chance of rain", "Peak 80% at 13:00") rendered fine but the Vico chart
+    // body didn't — see PR #400's flaky precipitation_card regen. This variant
+    // also retries if the bottom region of the snapshot is near-uniform, since
+    // chart cards always paint axis labels, grid lines and at least one data
+    // line across that area.
+    private fun captureChart(content: @Composable () -> Unit) {
+        composeRule.setContent { content() }
+        captureWithRetryIfEmpty(checker = ::chartLooksEmpty) {
+            composeRule.waitForIdle()
+            composeRule.onRoot().captureRoboImage(filePath = "$outputDir/${testName.methodName}.png")
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


mikelward:
Auto-rebase onto current `main` (post-#488 CI hardening) hit a conflict in `app/src/test/kotlin/app/clothescast/ui/today/PreviewSnapshots.kt`. Leaving the branch as-is; needs manual resolution before next CI run to pick up the new `ci.yml` (cleanup pass after the merge of #488).

---

PR #403, opened 2026-05-13
Show wind speed in mph when distance is set to miles

mikelward:
## Summary

The per-model wind diagnostic card hardcoded `km/h` in both its subtitle and Y-axis values, even for users on the imperial distance setting. This wires wind display through to `DistanceUnit` so:

- **km/h** is shown for users with distance set to kilometres
- **mph** is shown for users with distance set to miles

The same conversion also applies to the "Models disagree mostly… *X*–*Y* km/h" range shown under the forecast card when wind is the diverging factor (it's the same wind data, same screen).

No new user-facing setting — for now the wind unit is derived from the existing distance preference. A `TODO` on the new `WindSpeedUnit` enum flags promoting this to its own picker once knots (and possibly m/s) are added, since sailors / pilots will want wind unit chosen independently of distance.

### What changed

- new `WindSpeedUnit { KMH, MPH }` enum in `:core:domain` + `toWindSpeedUnit` / `symbol` / `DistanceUnit.windSpeedUnit()` helpers
- `TodayViewModel` now exposes `prefs.distanceUnit`; `WindCard` takes a `WindSpeedUnit` and converts both the picker values and the 10 km/h Y-axis floor (10 km/h ≈ 6.2 mph — kept as a converted floor so the "don't zoom calm days into noise" heuristic stays equivalent)
- `today_wind_subtitle` is now `Per-model wind speed (%1$s) by hour`
- new mph-variant preview + snapshot test pins the imperial render path; existing km/h snapshots are unaffected (substituted "km/h" matches the previous hardcoded copy byte-for-byte)
- domain tests for `toWindSpeedUnit`, `symbol`, and the derivation

### Privacy

No change to what leaves the device. Wind is still only rendered inside the app's diagnostic card — not in insight prose, not in TTS, not in clothes rules.

## Test plan

- [ ] JVM unit tests green (new `UnitsTest` cases for `toWindSpeedUnit`, `symbol`, and `DistanceUnit.windSpeedUnit()`)
- [ ] Android debug build green
- [ ] New Roborazzi snapshot `wind_card_with_model_spread_mph` lands; existing `wind_card_with_model_spread` is byte-identical
- [ ] Manual: with model-spread overlay enabled, flip distance unit between km / miles — wind card subtitle and Y-axis swap between "km/h" and "mph"; the divergence-summary range under the forecast card uses the same unit

---
_Generated by [Claude Code](https://claude.ai/code/session_01XjdBm4pnJEVgeavcxrtKS7)_


mikelward:
Snapshot regen at `f86a7a1` only added one image (existing km/h snapshots were byte-identical, as expected). Inlined for mobile reviewing:

**`wind_card_with_model_spread_mph.png`** (new)

wind_card_with_model_spread_mph


---

PR #421, opened 2026-05-13
Don't recompute today's nextOutfit from yesterday's tonight slot

mikelward:
## Summary

Follow-up to #419 addressing a [Codex P2](https://github.com/mikelward/clothescast/pull/419#pullrequestreview-…) that arrived after merge.

Morning settings edit scenario: the TODAY slot has just been rewritten by the morning worker (`forDate = today`), but the TONIGHT slot may still hold yesterday evening's insight (`forDate = yesterday`) because the evening worker hasn't fired yet. The recompute introduced in #419 unconditionally borrowed TONIGHT's hourly to re-derive TODAY's `nextOutfit`; when the dates didn't match, that silently rewrote today's "Tonight" icon and rationale from last night's overnight conditions — which have already happened.

## Fix

Guard the pairing on `forDate`: only borrow TONIGHT when its `forDate` matches TODAY's. Otherwise preserve the existing `nextOutfit` so the next TONIGHT worker run is what'll refresh it.

```kotlin
val tonightPair = tonightInsight?.takeIf { it.forDate == todayInsight?.forDate }
```

## Privacy

No off-device surface change — this is a pure local-cache correctness fix.

## Test plan

- [x] New `InsightCacheTest` case stores TODAY (today) + TONIGHT (yesterday) with deliberately different hourly profiles (today mild, yesterday overnight cold) and confirms `recomputeOutfits` leaves the TODAY slot's `nextOutfit` verbatim instead of rebuilding it from yesterday's data.
- [ ] CI: `:core:domain:test`, `:core:data:test`, `:app:testDebugUnitTest`, `:app:assembleDebug` (sandbox has no AGP / Android SDK; relying on CI per AGENTS.md).
- [ ] Manual: change a clothes setting in the morning before the tonight worker has run; verify the home-screen "Tonight" icon next to "Today" stays on the value the morning worker computed (rather than flipping to yesterday's overnight pick).

https://claude.ai/code/session_01P9K3fp87JFM8wnJmFgHMHn

---
_Generated by [Claude Code](https://claude.ai/code/session_01P9K3fp87JFM8wnJmFgHMHn)_


mikelward:
## UI snapshot diffs

Bot's "78 snapshots" was the run total; only 1 PNG actually differs, and it's unrelated to this PR.

### `precipitation_card_with_model_spread.png` — unrelated, ~0.01% byte diff

This PR is a data-layer-only change in `InsightCache.kt`; nothing in the precipitation card or model-spread overlay was touched. 60838 → 60845 bytes (7 bytes / 0.01%) is sub-pixel renderer noise, same kind of wobble the previous regen had on `wind_card_with_model_spread_mph.png`. Mentioning so it doesn't get blamed on the recompute fix.

Before (`7282daa`):
precipitation_card before

After (`9c09e83`):
precipitation_card after

---

PR #424, opened 2026-05-14
See each model's forecast by tapping any chart

mikelward:
## Summary

Replaces the persistent "show model spread" Display setting with a session-scoped tap-to-reveal gesture, and standardises every chart on the Today screen to draw a consensus main line by default with the per-model overlay shown only when the user taps to reveal it.

**Tap surfaces** — every chart card, plus the confidence chip and the low-confidence callout:

- Confidence chip + low-confidence callout grow a small "Tap to see each model's forecast" / "Tap to hide model forecasts" hint so the affordance is discoverable.
- Feels-like, air-temperature, precipitation cards.
- Wind, cloud, humidity, solar radiation, sunshine, UV index cards.

Every card now reads the same way: one consensus line by default, the spread underneath when toggled. The Y-axis stays sized to the per-model envelope regardless of toggle state, so axis labels don't shift under the data when you tap.

**Tap vs. scroll:** uses `Modifier.clickable {}` so a touch-and-drag that turns into a scroll won't fire the click — Compose's touch slop lets the parent scroll consume the gesture.

## What changes for users

- **Settings → Display → "Show model spread on charts"** is gone. The toggle lives on the Today screen now, scoped to the visit rather than persisted.
- **Diagnostic cards (wind, cloud, humidity, solar, sunshine, UV)** previously only rendered when the persisted setting was on, and only showed per-model lines. They now **always render** with a consensus main line, communicating the day's actual wind / cloud / sun / UV at a glance. Tapping reveals the per-model spread underneath.
- **Discoverability:** the confidence chip and low-confidence callout carry a "Tap to see each model's forecast" hint so users find the gesture without digging.

## Privacy

No off-device surface change. The model-spread toggle was always a local-display preference; removing its persistence layer doesn't change what crosses the device boundary. Per-model data is and has always been fetched on every cycle regardless of toggle state.

## Implementation notes

- `showModelSpread` removed from `UserPreferences`, `SettingsRepository`, `SettingsState`, `SettingsViewModel`, `SettingsScreen`, `DisplaySettings`. The DataStore key (`show_model_spread`) is no longer read or written; any persisted value is silently ignored.
- Old setting strings (`settings_display_charts_title`, `settings_display_model_spread_label`, `_description`) removed from the default locale and 39 translated locale files. New `today_confidence_tap_to_show` / `_hide` hint strings added to the default locale (other locales inherit until translated).
- `PerModelDiagnosticCard` grows two new params (`showOverlay`, `onToggleOverlay`) and computes a consensus main line in-card as a per-hour cross-model mean. Sparse handling carries through: a single missing hour drops the model from that hour's mean but doesn't drop the whole model from the chart.
- The session-scoped toggle lives as a `MutableStateFlow<Boolean>` on `TodayViewModel`, combined into `TodayState` alongside the existing prefs/work-status flows. Reset to off on process death by design — it's a "let me look at the spread right now" gesture, not a setting.

## Snapshots

- 3 new previews lock in the hint copy (`confidence_medium_tap_to_show`, `confidence_medium_tap_to_hide`, `low_confidence_callout_tap_to_show`).
- 6 new "consensus only" previews lock in the default appearance of each diagnostic card (`wind_card_consensus`, `cloud_card_consensus`, `humidity_card_consensus`, `solar_radiation_card_consensus`, `sunshine_card_consensus`, `uv_index_card_consensus`).
- The existing 6 "with model spread" diagnostic-card snapshots will change because the consensus main line now also draws over the per-model overlay.

## Test plan

- [ ] CI: `:core:domain:test` (`Settings.kt` shape compiles without `showModelSpread`)
- [ ] CI: `:app:testDebugUnitTest` covers the new + updated snapshots and the trimmed `SettingsRepositoryTest`
- [ ] CI: `:app:assembleDebug` builds with the locale-file string removals + new hint strings
- [ ] Manual on-device:
  - Tap the confidence chip → all 9 chart cards grow per-model overlays under their consensus lines; hint flips to "Tap to hide…"
  - Tap a chart card → overlay disappears, hint flips back
  - Tap-and-drag from inside a card → scroll happens, overlay does NOT toggle
  - Settings → Display: only the theme picker is left


mikelward:
### Snapshot diffs (regen `54275be`)

4 actual changes (3 added, 1 modified) out of the 84 the bot re-encoded.

#### Added — tap-hint variants

**Confidence chip · tap to show spread** (overlay is currently off, hint invites the user to enable it)
confidence_medium_tap_to_show

**Confidence chip · tap to hide spread** (overlay is currently on, hint flips to the inverse copy)
confidence_medium_tap_to_hide

**Low-confidence callout · tap to show spread**
low_confidence_callout_tap_to_show

#### Modified

**Settings → Display** — the "Charts" section + "Show model spread on charts" toggle row removed; only the theme picker is left.

Before (`ed981fe` on `main`):
settings_display before

After (`54275be`):
settings_display after

---

PR #425, opened 2026-05-14
Summarise wind, cloud and humidity at the top of each card

mikelward:
## Summary

The wind / cloud / humidity diagnostic cards' subtitles were static descriptions of what the chart plots ("Per-model wind speed (km/h) by hour"). Temperature already shows a min–max summary above its chart and precipitation a peak callout, so these three were the odd ones out — the user asked for something more useful there.

- **Wind** → peak consensus value + hour: `Peak 35 km/h at 16:00`. Gusts are the actionable signal, not the range.
- **Cloud** → min–max range of consensus: `Cloud 20 – 90%`.
- **Humidity** → min–max range: `Humidity 30 – 80%`.

The summary is computed from the same per-hour consensus blend the chart's main line draws (per-index mean across whichever models reported the metric at that hour), so the headline reflects what the chart shows. When no consulted model has data for the metric the card already auto-hides via `PerModelDiagnosticCard`'s early return; the subtitle falls back to the previous descriptive text as a safety net.

## Changes

- `WindCard` / `CloudCard` / `HumidityCard` in `TodayScreen.kt`: compute a summary from the per-model consensus series and feed it to `PerModelDiagnosticCard.subtitle`.
- New private helpers `perModelConsensusSeries` and `perModelConsensusRange` next to `formatMinMax` — they mirror `PerModelDiagnosticCard`'s existing main-line blend.
- New strings: `today_wind_peak`, `today_cloud_range`, `today_humidity_range`.

## Privacy

No change to what leaves the device — these are purely screen-side summaries computed from data already fetched.

## Test plan

- [ ] CI: `JVM unit tests` (Roborazzi snapshots will need regenerating — the wind / cloud / humidity card snapshots now show the computed subtitle instead of the static one).
- [ ] CI: `Android debug build`.
- [ ] Manual: load the Today screen on a day where a peak hour is well-defined and confirm wind subtitle reads "Peak N km/h at HH:00"; flip distance unit to imperial and confirm it re-renders in mph.
- [ ] Manual: confirm cloud / humidity subtitles show min–max values matching the chart bounds.
- [ ] Manual: on a build with no per-model data (older cache payload), the cards still auto-hide as before.

I couldn't run the build locally — Google Maven is blocked in this sandbox so AGP won't load. CI is the validation surface.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VyazPYxqcsdSfLKKLmMEx8)_


mikelward:
Snapshot diffs from `e8d9359`'s regen — embedded inline so they're visible from the mobile app (the Files tab shows "Binary files not rendered"). Before = `bf094bf` (parent, pre-regen), after = `e8d9359`.

### Wind

`wind_card_consensus` — subtitle is now `Peak N km/h at HH:00` instead of the static description.

Before (`bf094bf`):
wind_card_consensus before
After (`e8d9359`):
wind_card_consensus after

`wind_card_with_model_spread`:

Before:
wind_card_with_model_spread before
After:
wind_card_with_model_spread after

`wind_card_with_model_spread_mph`:

Before:
wind_card_with_model_spread_mph before
After:
wind_card_with_model_spread_mph after

`wind_card_sparse_trailing` — peak-and-time still gets a sensible value even when the leading hours are missing for some models.

Before:
wind_card_sparse_trailing before
After:
wind_card_sparse_trailing after

### Cloud

`cloud_card_consensus` — subtitle is now `Cloud N – N%` instead of the static description.

Before:
cloud_card_consensus before
After:
cloud_card_consensus after

`cloud_card_with_model_spread`:

Before:
cloud_card_with_model_spread before
After:
cloud_card_with_model_spread after

### Humidity

`humidity_card_consensus` — subtitle is now `Humidity N – N%`.

Before:
humidity_card_consensus before
After:
humidity_card_consensus after

`humidity_card_with_model_spread`:

Before:
humidity_card_with_model_spread before
After:
humidity_card_with_model_spread after

### Precipitation (unexpected — looks like Vico rendering noise)

`precipitation_card_with_model_spread` shifted too even though this PR doesn't touch the precipitation card or its data sources. The subtitle is unchanged (`Peak 80% at 13:00`) and only this card renders in the preview. Prior regen (`1bb464f`) showed the same pattern — unrelated snapshots nudged by a few bytes — so I'm reading this as Vico/Roborazzi chart-rendering nondeterminism rather than a real regression. Heads-up though, worth a glance:

Before:
precipitation_card_with_model_spread before
After:
precipitation_card_with_model_spread after


---

PR #428, opened 2026-05-14
Add a colour-blind friendly palette for charts and confidence cards

mikelward:
## Summary
Adds a **Palette** picker on the Display settings page with two named choices:

- **Rainbow** (default) — the saturated pink (ECMWF) / orange (GFS) / green (ICON) model trio and Material teal/red confidence card the app shipped with.
- **Accessible** — an Okabe-Ito-derived trio (vermillion / orange / bluish-green plus a darker grey for the best-match overlay) and sky blue / neutral / amber confidence backgrounds. Designed to stay distinguishable under deuteranopia, protanopia, and tritanopia; the dark-mode variant flips background and foreground so the card stays legible in both themes.

The "Rainbow ↔ Accessible" framing is deliberate — it's a palette pick, not an accommodation. Everyone sees both options; the explainer text under each option mentions the colour-blindness use case so users know what the trade-off is.

## Why
The default Rainbow trio pairs pink (`#D81B60`) with green (`#43A047`), which collide under deuteranopia and protanopia (~5% of men). The confidence card's HIGH-vs-LOW contrast leans on Material's teal `secondaryContainer` vs. red `errorContainer` — the textbook red-green trap. Text labels still disambiguate, but the colour cue does little work for those users. The Accessible palette swaps both at once so the hue itself reinforces the label rather than fighting it.

## Implementation notes
- New `AppPalette` carried through a `LocalAppPalette` `CompositionLocal`. `ClothesCastTheme` resolves and provides it from the user's `ColorPalette` choice + the active `ColorScheme`. Every previous reference to `secondaryContainer` / `errorContainer` on `ConfidenceChip` and to the hard-coded `MODEL_COLORS` map in the chart layer / `ModelSpreadLegend` now reads from the palette, so the two palettes swap every consumer in lockstep.
- `ColorPalette { RAINBOW, ACCESSIBLE }` lives in `:core:domain` so it's testable without the Android SDK. Persisted in DataStore under `color_palette` (string-stored enum, defaulting to `RAINBOW`).
- `MainActivity` reads the pref synchronously on cold start (same flicker-avoidance trick as theme mode) and observes the flow at runtime so a settings flip takes effect immediately without an Activity recreate.
- The local sandbox couldn't reach Google Maven, so `:app:assembleDebug` / `:app:testDebugUnitTest` weren't runnable here. Relying on CI for Android-side validation, per CLAUDE.md.

## Privacy
No change to what leaves the device. Palette is render-time only — not in any insight prose, TTS preamble, Firebase payload, or weather request.

## Test plan
- [ ] CI: `JVM unit tests` (covers the new `AppPaletteTest` + the `SettingsRepository` round-trip for `colorPalette`)
- [ ] CI: `Android debug build` (the only Android-side smoke this PR needs, since palette is render-time)
- [ ] Roborazzi snapshot regen — the `SettingsDisplayPreview` will gain the new "Palette" section card and CI will commit the regenerated PNG
- [ ] Manual: open Display settings → pick **Accessible** → verify per-model chart lines, model spread legend, and confidence card all flip together in both light and dark themes
- [ ] Manual: kill and relaunch the app → verify the chosen palette survives a cold start and the first frame already matches the pick (no flicker through Rainbow)

https://claude.ai/code/session_011bpFZ99owGSabb9bn1ieuW


mikelward:
Snapshot diffs from `ab2b303`. Only 2 PNGs actually changed vs `main` (the bot's "88 regenerated" counts total files in `app/snapshots/`, not the delta):

### `settings_display.png` — Display settings, new Palette section card

**Before** (`fbd95f7`):

settings_display before

**After** (`ab2b303`):

settings_display after

### `precipitation_card_with_model_spread.png` — Precip card with per-model overlay

**Before** (`fbd95f7`):

precipitation before

**After** (`ab2b303`):

precipitation after

The Rainbow palette returns the same hex values as the old `MODEL_COLORS` map, so the precip card pixels should be identical — happy to dig in if the diff is visible. The forecast card with the same overlay pattern didn't shift, which suggests it's either render order or anti-aliasing rather than a colour drift.

---

mikelward:
New accessible-palette snapshots from regen `489d370`:

### `forecast_card_with_model_spread_accessible.png` — Forecast card with model spread, Accessible palette

forecast accessible

### `precipitation_card_with_model_spread_accessible.png` — Precip card with model spread, Accessible palette

precipitation accessible

For comparison, the Rainbow versions are in the [earlier diff comment](https://github.com/mikelward/clothescast/pull/428#issuecomment-4449368507). The Accessible palette swaps ECMWF pink → vermillion (`#D55E00`), GFS orange → richer orange (`#E69F00`), ICON green → bluish-green (`#009E73`); the blended main line stays theme primary blue.

---

PR #479, opened 2026-05-15
Show BOM in the model picker, disabled, while it's under maintenance

mikelward:
## Summary

The Australian BOM model has been quietly missing from Settings ▸ Forecasters since Open-Meteo's BOM open-data delivery was suspended for their platform upgrade — users hunting for it had no signal that it was a known gap rather than an oversight on our end.

This PR surfaces BOM as a permanently-disabled row at the bottom of the picker (geographically after JMA), with the subtitle:

> Australian — under maintenance, back as soon as possible.

so it's visibly known-missing.

## Scope

- The row is cosmetic-only. BOM stays out of `ForecastModel`, so no fetch / chart / palette / persistence path touches it. The checkbox is permanently unchecked, `enabled = false`, and the `onToggle` callback is a no-op.
- The `Settings.kt` enum comment is updated to point at the new picker placeholder; the `strings.xml` comment is updated to match. When BOM delivery resumes, the row should be removed and BOM wired through the enum proper.
- Adds a `settings_forecasters` Roborazzi snapshot (Custom mode, anchored to London, four models selected) so the disabled BOM row is visible in PR review and any future churn surfaces as a visual diff. The picker is taller than the suite's class-level 640dp viewport, so the new test overrides `@Config` qualifiers to 1024dp tall.

## Privacy

No change to what leaves the device — the row is render-only and never participates in a network request.

## Test plan

- [x] `:app:testDebugUnitTest` passes locally (all 422 tests; previous CI flakes were Maven/network-side, unrelated).
- [x] `:app:assembleDebug` compiles.
- [x] New `settings_forecasters.png` snapshot recorded and committed; `no_preview_without_snapshot` meta-test still passes.
- [ ] Manual: open Settings ▸ Forecasters and confirm the BOM row appears at the bottom, greyed out, with the maintenance subtitle, and that tapping it does nothing.
- [ ] Manual: toggle Auto on/off and confirm the BOM row stays disabled in both modes.


mikelward:
New `settings_forecasters` snapshot — Custom mode, anchored to London with ECMWF/GFS/ICON/UKMO selected so the checked / unchecked-but-enabled / permanently-disabled rows are all in frame. The BOM row at the bottom is the placeholder this PR adds:

settings_forecasters


---

PR #530, opened 2026-05-17
Soft fade hints when there's more to scroll on Today

mikelward:
## Summary

The Today screen's pager pages each scroll vertically through outfit row → insight card → confidence chip → forecast + precipitation charts → six per-model diagnostic cards. Right now there's no visual signal that there's anything below the fold, so a user who doesn't habitually scroll never discovers the forecast. `942a9165` made the confidence chip tap-to-scroll, but that only helps users who tap.

Adds a soft background-coloured gradient at the top and bottom edges of each pager page's viewport:

- **Bottom fade** visible whenever there's more content below.
- **Top fade** visible once scrolled past the start.
- Each animates in / out via `animateFloatAsState` driven by `scrollState.canScrollForward` / `canScrollBackward`, so the page lands in the correct state on first paint and transitions smoothly thereafter.

The horizontal axis is **unchanged** — the existing chevrons in `InsightCard` (`TodayScreen.kt:1306–1319`) already signal "more sideways", and a second chevron at the page bottom would compete with them. Top / bottom (not start / end) keeps the fades RTL-neutral.

## Implementation notes

- New `EdgeFadeOverlay` internal composable in `TodayScreen.kt` wraps its content in a `Box` and draws two 32 dp `Brush.verticalGradient` bands aligned to the top and bottom edges. Visibility is driven by two `Boolean` flags so the composable is testable from previews that drive the alphas directly.
- `TodayPage` calls it with `scrollState.canScrollBackward` / `canScrollForward`.
- The overlays sit at the pager page's outer bounds (not inside the Column's internal `padding(top = 16.dp, bottom = 24.dp)` from `fc37b9cd`). Cards pass cleanly under the fades.
- Each overlay is a plain `Box` with `Brush.verticalGradient` + `Modifier.alpha`. No `clickable`, so drag gestures fall through to the underlying scroll and the pager swipe still works.
- Fade colour is `colorScheme.background` — same colour the Scaffold uses as its default `containerColor`, so the gradient reads as content fading into the screen background.
- Each of the two pager pages owns its own `scrollState`, so the fades animate independently per page.

## Snapshot coverage

Four new Roborazzi PNGs under `app/snapshots/` lock in the affordance's visible states (top-only / both / bottom-only) plus a dark-theme variant against stand-in card content, so any regression in fade height, gradient direction, alpha animation, or background-colour blend surfaces as a visible pixel diff:

- `edge_fade_bottom_only.png`
- `edge_fade_both.png`
- `edge_fade_top_only.png`
- `edge_fade_both_dark.png`

(Inline rendered diff posted as a comment on the PR.)

## Privacy

No change. Fades are local UI; nothing crosses the device boundary.

## Test plan

- [x] `./gradlew :app:compileDebugKotlin` clean.
- [x] `./gradlew :core:domain:test :core:data:test :app:testDebugUnitTest` green.
- [x] `./gradlew :app:assembleDebug` succeeds.
- [x] New `edge_fade_*` PNGs captured under `app/snapshots/`; no other PNGs changed.
- [ ] CI green (both jobs).
- [ ] Manual: install debug APK and verify on Today screen:
  - At rest (top): only bottom fade visible.
  - Mid-scroll: both fades visible.
  - At bottom: only top fade visible.
  - Swipe sideways: each page animates its own fades; horizontal chevron still works.
  - Drag through a fade region: scroll continues — fade doesn't swallow gestures.


mikelward:
Snapshot diffs for the new edge-fade affordance — four added PNGs (added files, so just the new state). All four use stand-in cards sized to a fixed 400 dp region so the 32 dp gradient bands sit at known offsets.

**At top — bottom fade only** (`edge_fade_bottom_only.png`)
edge_fade_bottom_only

**Mid-scroll — both fades** (`edge_fade_both.png`)
edge_fade_both

**At bottom — top fade only** (`edge_fade_top_only.png`)
edge_fade_top_only

**Dark theme, both fades** (`edge_fade_both_dark.png`)
edge_fade_both_dark

---

PR #554, opened 2026-05-17
Toggle global holidays on or off as their own bucket

mikelward:
## Summary

Promotes the four universal holidays (Christmas, New Year's,
Valentine's Day, Halloween) from "fold into every country's bucket"
to a first-class country-like entity in Settings → Holidays.

- New `Global` checkbox in the top card, sitting between Current and
  All, peer to Home / Current. Greyed out when All is on, same as
  the other two non-All toggles.
- New `Global` expandable section above the alphabetised ISO country
  list, listing the four universal holidays with the same `Auto / On
  / Off` country-override dropdown that France and Japan already
  have. Backed by the existing per-country override path keyed by
  `HolidayCatalog.GLOBAL_COUNTRY` — no new persistence machinery.
- Per-country expandables now show only that country's holidays.
  Opening "France" no longer duplicates Christmas, NYE, etc.; those
  live in the dedicated Global section.
- Domain: `HolidayCountrySelection.global: Boolean = true` joins
  `home`/`current`/`all`. `countryAutoEffective` returns
  `all || global` for `GLOBAL_COUNTRY`. `resolveEnabledCountries`
  drops the post-loop "if non-empty, add GLOBAL" fixup — the global
  bucket now flows through the same code path as ISO codes.
- Persistence: new `holiday_country_global` boolean key in DataStore.
  Missing-key reads as `true`, so existing installs keep their
  universal-holiday themes firing after upgrade.

## Test plan

- [x] `cd core && ../gradlew :core:domain:test` — green
- [x] `./gradlew :core:domain:test :core:data:test :app:testDebugUnitTest` — green
- [x] `./gradlew :app:assembleDebug` — green
- [x] Roborazzi regen of `settings_holidays.png` shows the new
      checkbox row and Global section above the country list
- [ ] Manual smoke on device: untick Global; verify French / Japanese
      per-country expandables no longer show Christmas / NYE /
      Valentine's / Halloween; verify Today on Dec 25 no longer
      themes the outfit with Global off

## Privacy

No new off-device payloads. The `global` flag is a local preference;
rendered insight prose composition is unchanged beyond which
holidays surface.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WhtoSS66ZnEbSpQU7KcMiR)_


mikelward:
Snapshot diff — `app/snapshots/settings_holidays.png`:

Before (`8b021b4`):

settings_holidays before 8b021b4

After (`8d6980e`):

settings_holidays after 8d6980e

The new render shows the four-checkbox top card (Home / Current location / Global / All), a dedicated `Global 4/4 Auto (on)` section sitting directly above the alphabetised ISO countries, and per-country buckets that no longer fold in the universal holidays (e.g. Australia drops from `10/10` to `6/6`, Brazil from `1/5` to `0/1`).


---

PR #590, opened 2026-05-19
Show the outfit image full-screen on smart displays

mikelward:
## Summary

Before, the cast load handed the Default Media Receiver an `audio/wav` stream with the outfit PNG attached as a `WebImage`. DMR renders that as an "audio with album art" experience — the image is framed by the title / subtitle / progress chrome rather than filling the screen.

This PR muxes the outfit PNG (as a static H.264 video track at 1 fps) and the TTS audio (re-encoded to AAC LC) into a single MP4 served at the same token-gated LAN endpoint. `video/mp4` content renders the video track full-bleed on DMR, so the outfit fills the smart display while the forecast plays.

## Implementation

- **`Mp4Encoder.kt`** (new) — uses `MediaCodec` (H.264 + AAC) + `MediaMuxer` to produce the MP4 in memory. Concurrent encode loop pumps both codecs, buffering any early samples until both tracks are added to the muxer. ARGB→I420 conversion handles both packed (I420) and interleaved (NV12/NV21) encoder input formats via plane strides. Bitmaps are capped at 1280px on the long edge and rounded to even dimensions so weak hardware encoders accept them.
- **`CastMediaServer.kt`** — now serves a single `/{token}/insight.mp4` endpoint instead of separate `insight.wav` + `outfit.png` endpoints. Same token-gating, same LAN-only threat model.
- **`CastInsightController.kt`** — calls `Mp4Encoder.encode(png, wav)` on `Dispatchers.Default` before publishing, builds a `MEDIA_TYPE_MOVIE` / `video/mp4` `MediaInfo` instead of `MEDIA_TYPE_GENERIC` / `audio/wav` + `WebImage`.
- **`silentWavStub`** grows from 0.5 s to 15 s so the image-only fallback (Gemini unavailable or synth failed) still has reasonable on-screen dwell time when there's no spoken forecast to carry the duration.

## Privacy

No change to what leaves the device. The same outfit PNG and the same TTS audio that were already cast are now packaged into a single MP4 for the same Cast receiver; nothing new is sent, logged, or persisted.

## Test plan

- [x] `:app:testDebugUnitTest` — `CastMediaServerTest` updated for the single MP4 endpoint, all cast tests pass.
- [x] `:app:assembleDebug` — clean.
- [ ] On-device verification: cast a morning forecast to a Nest Hub / Chromecast and confirm the outfit fills the screen while the audio plays (full-bleed video, no audio-player chrome).
- [ ] On-device image-only path: temporarily break Gemini TTS, confirm the outfit still shows for ~15 s with no audio.
- [ ] Confirm the MP4 round-trips through both packed-plane (Pixel / Qualcomm) and interleaved-plane (some MediaTek) encoders without colour artefacts — the ARGB→I420 path takes both routes.

https://claude.ai/code/session_01VARNXKVqVSKPGqUWfP7Ucw

---
_Generated by [Claude Code](https://claude.ai/code/session_01VARNXKVqVSKPGqUWfP7Ucw)_


mikelward:
Snapshot regen touched one file — `app/snapshots/settings_smart_home.png`. The diff is byte-level encoder noise (225 859 → 225 408 bytes, ~0.2%); this PR doesn't touch any UI code (`git diff` is cast/server/test only), so this is environmental drift, not a real visual change.

**Before** (`adbe212` on main):

settings_smart_home @ adbe212

**After** (`ec713b8` on this PR):

settings_smart_home @ ec713b8

---

PR #611, opened 2026-05-21
Grant calendar permission from the toggle, not a separate button

mikelward:
## Summary

Follow-up to #610. Simplifies the Calendar settings permission UI:

- **Removes the "Grant calendar permission" button.** The "Use my calendar" toggle already fires the `READ_CALENDAR` prompt when turned on, so the button was redundant.
- **Adds an "Open system settings" link, shown only when the permission is permanently denied** ("don't ask again"). That's the one state the toggle can't recover from — Android stops showing the runtime dialog, so the only path back is the app's system-settings screen. Detected via `shouldShowRequestPermissionRationale` returning false after a denial.
- **Turning the master off** just stops calendar reads in-app (the `*Active` gates). We deliberately **don't** relinquish the OS permission — there's no immediate API, and `revokeSelfPermissionsOnKill` is API 33+ (minSdk is 31) and only takes effect after the process is killed; a code comment records why.
- **Renames the card heading** to "Calendar permission" (matches the toggle's intent).

## Test plan

- [x] `:core:domain:test` + `:app:testDebugUnitTest` — green; `settings_calendar.png` re-recorded (no grant button; "Calendar permission" heading).
- [x] `:app:assembleDebug` — green.
- [ ] On-device: deny calendar twice → confirm "Open system settings" link appears; grant from settings → link clears and toggle reflects granted.

## Notes
The permanently-denied "Open system settings" affordance is local composable state, so it isn't captured in the snapshot (the recorded state is the normal, not-yet-denied card).

https://claude.ai/code/session_018FhmCHA22Rnd2Sgmd27uhv

---
_Generated by [Claude Code](https://claude.ai/code/session_018FhmCHA22Rnd2Sgmd27uhv)_


mikelward:
**Snapshot diff** — `settings_calendar.png` (the only changed snapshot).

Heading renamed `Calendar access` → **Calendar permission**, and the always-shown "permission denied" text + "Grant calendar permission" button are gone (the toggle handles the prompt). The "Open system settings" link only shows on permanent denial, which isn't reachable from a preview, so it's not in the recorded state.

**Before** (`9ab351e5`, main):

before

**After** (`41e89995`):

after

---

PR #615, opened 2026-05-21
Choose what the ClothesCast says with new Format settings

mikelward:
## Summary

Adds a **Format** settings page (subhead **"What to say"**) that lets the user configure how each clause of the daily insight prose reads, via a dropdown per clause:

- **Temperature range** — None / Degrees / Bands. "Bands" renders the existing `TemperatureBand` classification as words (e.g. "cool to mild") using the already-translated `insight_band_*` strings that were defined but never wired up. "None" generalises the previous omit-temperature-range toggle.
- **Significant change** — Off, or a preset °C threshold (3 / 5 / 8), replacing the hard-coded `3.0`°C gate in `RenderInsightSummary`. Stored as °C and labelled in the user's temperature unit (Fahrenheit users see 5 / 9 / 14) so it stays consistent with the rendered "5° warmer than yesterday." delta.

The page is reachable three ways:
- Settings root menu, directly below **Clothes**.
- A small **Format** link in the insight card's bottom-right corner on the Today screen.
- **Long-pressing the prose body** of the insight card (just the prose — not the date/location header or the generated-at footer).

It shows a **live preview** that re-renders sample prose as the settings change.

## Details

- The old `omitTemperatureRange: Boolean` preference is replaced by a `RangeFormat { NONE, DEGREES, BANDS }` enum plus a nullable `deltaThresholdC` on `UserPreferences`. The omit-range toggle is removed from the Schedule page.
- **Migration:** existing installs with `omit_temperature_range = true` map to `RangeFormat.NONE` on read; an absent delta-threshold key keeps the historical 3°C default.
- `RenderInsightSummary.invoke(...)` gains a `deltaThresholdC` parameter (defaulting to 3.0 so existing callers/tests are unchanged); `GenerateDailyInsight` threads `prefs.deltaThresholdC` through.

## Privacy

No change to what leaves the device — this only reshapes already-rendered prose, which still feeds TTS / MQTT exactly as before. The Firebase `settings_snapshot` event swaps its `omit_temperature_range` flag for `range_format` (enum name) + `delta_threshold_c` (integer °C, -1 = off); no user content. The telemetry privacy-contract test is updated to match.

## Test plan

- [x] `:core:domain:test` — added cases for threshold off (null) and a custom threshold above the delta.
- [x] `:app:testDebugUnitTest` — updated `InsightFormatterTest` (NONE + new BANDS rendering tests) and `TelemetryPrivacyContractTest`.
- [x] `:app:assembleDebug` passes.
- [x] New `settings_format` Roborazzi snapshot added; no other snapshots changed.
- [ ] Manual: long-press / corner link opens Format; switching Range to Bands shows "cool to mild"; Significant change Off hides the delta clause; higher threshold suppresses a small delta.

https://claude.ai/code/session_011ugRS6bxvXj3ag7y1CEEuH

---
_Generated by [Claude Code](https://claude.ai/code/session_011ugRS6bxvXj3ag7y1CEEuH)_


mikelward:
New UI snapshot for the Format settings page (added — inline so it renders on mobile, where the Files tab shows "Binary files not rendered"):

settings_format

The live preview at the top re-renders the sample prose as the dropdowns change; here Range = Bands ("cool to mild") and Significant change = 3°.

---

PR #629, opened 2026-05-22
Preview your current forecast on the Format screen

mikelward:
## Summary

The Format settings page now shows **two** preview cards instead of one:

- **Example** — the existing synthetic preview (renamed from "Preview"), still showing every clause with sample values so the user can see how each format setting reads.
- **Current forecast** — a new card that renders the user's *real* cached current insight (page 1 of the Today pager) through the same `InsightFormatter` the Today screen uses. Falls back to "No forecast yet — check back after your next ClothesCast." when nothing's cached.

The range-format setting updates the current-forecast card live. The delta-threshold and clothes-mention settings are baked into the cached summary at generation time, so they re-take effect on the next refresh — same behaviour as the Today screen, which also formats the cached summary as-is.

## Implementation

- `SettingsViewModel` now takes an optional `InsightCache`, observes `thisPeriod`, and exposes the summary via the new `SettingsState.currentInsightSummary`. Defaulted to `null` so pure-VM tests are unaffected; the Activity wires the real cache through the nav host.
- `FormatContent` gains an optional `currentInsightSummary` param and a new `CurrentForecastPreviewCard` composable.
- Renamed the example card's title string and added strings for the new card + empty state.

## Privacy

No change to what leaves the device. The new card reads the on-device cached `InsightSummary` only — nothing new is sent off device.

## Test plan

- [x] `:app:testDebugUnitTest` passes locally (Roborazzi snapshots regenerated).
- [x] New snapshot `settings_format_current_forecast.png` shows both populated cards; existing `settings_format.png` / `settings_format_clothes_never.png` updated to show the empty-state card.
- [ ] CI green.

https://claude.ai/code/session_01Uo9663Q22xpTxzSpfr81U2

---
_Generated by [Claude Code](https://claude.ai/code/session_01Uo9663Q22xpTxzSpfr81U2)_


mikelward:
Snapshot changes (GitHub's mobile app shows "Binary files not rendered" for PNG diffs, so inlining them here):

### `settings_format_current_forecast.png` (new)
The two-card layout with a populated current forecast.

new

### `settings_format.png` (modified) — now shows the empty-state current-forecast card

Before (`c269b46`):

before

After (`20f2cfa`):

after

### `settings_format_clothes_never.png` (modified) — same, with clothes-mention NEVER

Before (`c269b46`):

before

After (`20f2cfa`):

after

---

PR #632, opened 2026-05-23
Pick your fallback top and bottom for when no rules match

mikelward:
## Summary

- Renames the "Standard bottom" Settings card to **"If no rules match"** and expands it to two rule-row-style edit rows: one for the fallback top (default T-shirt) and one for the fallback bottom (default Long pants).
- Each row shows the temperature range the fallback would apply at, derived from your current rules and rendered in your unit ("when temperature is above 18°C"). When rules cover everything, the row reads "never"; with no relevant rules, the secondary line is omitted.
- Adds `defaultTop` to `UserPreferences` / DataStore; threads it through `OutfitSuggestion.fromForecast`, the insight cache, and the home-screen widget refresh path so picking a new fallback flips the icon immediately. `defaultBottom` validation now accepts every `Bottom` (including SHORTS for hot-climate users who removed the shorts rule).
- New domain helper `fallbackRange(rules, tier)` in `:core:domain` computes the un-covered temperature window from the rules list — single source of truth for the row's secondary description.

## Privacy

`SettingsSnapshot` gains one new field, `defaultTop` (analytics event param `default_top`). It's the enum name only (e.g. `TSHIRT`, `POLO`) — same shape and category as the existing `default_bottom` param. The privacy-contract test was updated to include it in the allowlist.

The old `settings_default_bottom_title` / `_description` strings were removed from `values/` and the 44 locale overlays — they're no longer referenced. The new English strings (`settings_default_outfit_*`) ship un-translated; other locales fall back to English until the translation pass.

## Test plan

- [x] `:core:domain:test` — new `FallbackRangeTest` covers default rules, empty list, conflicting bounds (`empty=true`), Fahrenheit normalisation, and the "warmest threshold wins" rule for multiple `TemperatureBelow` entries. `OutfitSuggestionTest` gains two `defaultTop` round-trip cases.
- [x] `:core:data:test` — passes.
- [x] `:app:testDebugUnitTest` — `SettingsRepositoryTest` gains `defaultTop` round-trip + unknown-value-clamps-to-TSHIRT cases; the relaxed `defaultBottom` validation (SHORTS allowed) is also covered. `InsightCacheTest` updated for the new `recomputeOutfits` signature. `TelemetryPrivacyContractTest` allowlist updated.
- [x] `:app:assembleDebug` — passes.
- [x] Roborazzi snapshots regenerated (`settings_clothes.png` / `settings_clothes_fahrenheit.png`) — the new section is visible and renders the rule-row pattern.

https://claude.ai/code/session_01VNqpSkA6FJPArac4hDGrQV

---
_Generated by [Claude Code](https://claude.ai/code/session_01VNqpSkA6FJPArac4hDGrQV)_


mikelward:
Snapshot diffs inline for the mobile app (GitHub's Files tab renders these PNGs as "Binary files not rendered").

**`settings_clothes.png` (Celsius)** — `8f03aae` → `c5d3a4d`

Before — radio buttons under "Standard bottom":

settings_clothes before

After — edit rows under "If no rules match" with derived temperature ranges:

settings_clothes after

**`settings_clothes_fahrenheit.png` (Fahrenheit)** — `8f03aae` → `c5d3a4d`

The Fahrenheit fixture has an extra user-added rule on top of the defaults, so the new section sits below the viewport here — but you can still see the "If no rules match" header at the bottom of the after, and the "T-shirt" row's name peeking through.

Before:

settings_clothes_fahrenheit before

After:

settings_clothes_fahrenheit after

---

PR #670, opened 2026-05-24
Move the chart time readout out of a tooltip into each card

mikelward:
## Summary

Follow-up to #668 — moves the indicator-time readout out of the floating tooltip and into each card's existing text column.

- Each chart card renders the readout (e.g. `14:30 · 18°C`) as an extra line below its current subtitle. Visible whenever the shared `ChartScrubController` has an active time in this chart's window, hidden otherwise (Tomorrow page before first tap).
- The floating Surface that previously sat on top of the chart canvas is removed. `ChartScrubOverlay` becomes `ChartRestoreOverlay` and just hosts the top-right restore icon (still an interactive control, so it stays on the chart).
- Shortens two diagnostic subtitles so the new readout line sits cleanly under them: `today_solar_subtitle` "Surface irradiance (W/m²) by hour" → "Surface irradiance (W/m²)"; `today_uv_subtitle` "UV index by hour" → "UV index". Both lines ship through the Play "What's new" → so they're written as user-facing copy.

Snapshots regenerated:
- `forecast_chart_with_current_time.png` — the `ForecastChartWithCurrentTimePreview` no longer shows the floating tooltip.
- `solar_radiation_card_consensus.png` and `solar_radiation_card_with_model_spread.png` — shortened solar subtitle.

## Privacy

No change — readout text renders existing values from the in-memory insight cache.

## Test plan

- [ ] `./gradlew :app:testDebugUnitTest` — passes locally; the three snapshot deltas are intentional and committed.
- [ ] Today screen: each card's text column reads "TIME · VALUE" at the indicator. Updates as the minute ticks. Scrubbing on any chart updates the readout on every card simultaneously.
- [ ] Restore icon still appears top-right of chart canvas only while scrubbed; tap returns to "now"; readout follows back.
- [ ] Tomorrow page: no readout line at idle (no `currentTimeChartX` match); appears after first tap, hides on restore.
- [ ] Light + dark themes — readout text uses theme typography (bodyMedium), no contrast issues.

---
_Generated by [Claude Code](https://claude.ai/code/session_01J3GLT6Vxsrc7vU83RRrcoX)_


mikelward:
Screenshot diffs for the three regenerated snapshots — embedded inline so they render in the GitHub mobile app (binary file diffs don't).

**`app/snapshots/forecast_chart_with_current_time.png`** — floating tooltip removed; vertical now-indicator unchanged.

Before (a3fa63a5):
before

After (760ccdbd):
after

**`app/snapshots/solar_radiation_card_consensus.png`** — shortened subtitle ("Surface irradiance (W/m²) by hour" → "Surface irradiance (W/m²)").

Before (a3fa63a5):
before

After (760ccdbd):
after

**`app/snapshots/solar_radiation_card_with_model_spread.png`** — same subtitle change.

Before (a3fa63a5):
before

After (760ccdbd):
after

The `ForecastChartWithCurrentTimePreview` only exercises the now-indicator branch (no scrubbed state), so we don't yet have a snapshot that shows the new readout-in-subtitle line. Worth adding a preview that drives the controller via `scrubTo(...)` to lock that path into a snapshot too — happy to throw it in if you want.

---

PR #671, opened 2026-05-24
Move chart restore icon to the card corner and inline the scrub readout

mikelward:
## Summary

Three paired changes:

- **Restore icon moves to the card corner.** The little clock-back icon used to sit at the chart canvas's top-right, where it overlapped data lines on busy charts. It now lives in the top-right corner of each chart *card* — outside the canvas, above the title row — so the plot stays uncluttered. Each card wraps its existing `Column` in a `Box` so `ChartRestoreOverlay` can align to the card's `TopEnd`.
- **Readout appends to the subtitle on the same line.** The indicator-time readout (e.g. `14:30 · 16°C`) now joins the card's existing subtitle as `"<subtitle> · <readout>"` instead of rendering as a separate row below. Reverts a wrong call in #670 — intended placement was always the existing subtitle line. `ChartReadout` becomes `rememberChartReadout` returning a `String?` the card splices in via `appendReadout`.
- **New scrubbed-state snapshot tests.** `ForecastCardScrubbedPreview` and `WindCardScrubbedPreview` drive a card through the shared `ChartScrubController` at 14:30, locking the appended subtitle, scrubbed indicator, and new restore-icon position into snapshots. Wind covers the `rememberChartReadout` + `tooltipValueFormat` path; Forecast covers the inline temperature-readout path.

## Privacy

No change.

## Test plan

- [ ] `./gradlew :app:testDebugUnitTest` — passes locally; two new PNGs (`forecast_card_scrubbed.png`, `wind_card_scrubbed.png`) committed.
- [ ] Today screen: restore icon now appears top-right of the card (not the chart) when scrubbed; tap snaps back to "now".
- [ ] Subtitle reads "<existing peak/range> · TIME · VALUE" while the indicator has an active time; same line; reverts cleanly to the existing peak/range text when no controller / Tomorrow page idle.
- [ ] Diagnostic cards (Wind/Cloud/Humidity/Solar/Sunshine/UV): same — appended subtitle + icon at card corner.


mikelward:
Scrubbed-state snapshots — new this PR, so no "before" to compare against. Embedding inline since binary file diffs don't render in the mobile app.

**`app/snapshots/forecast_card_scrubbed.png`** — temperature card, inline-readout path.

forecast_card_scrubbed

**`app/snapshots/wind_card_scrubbed.png`** — diagnostic card, `rememberChartReadout` + `tooltipValueFormat` path.

wind_card_scrubbed

Both show the appended subtitle ("<peak/range> · 14:30 · VALUE"), the indicator pinned at 14:30, and the restore icon at the card's top-right corner.

---

PR #673, opened 2026-05-24
Match the chart scrub readout's format to the existing card subtitle

mikelward:
## Summary

Follow-up to #671. Three consistency fixes so the appended readout reads as part of the same sentence as the existing subtitle:

- **Word order.** Swap "TIME · VALUE" → "VALUE at TIME" via a new `today_chart_readout` string resource (`"%1$s at %2$s"`). Mirrors the existing `today_wind_peak` / `today_uv_peak` / etc. patterns, so the appended subtitle now reads `Peak 16 km/h at 18:00 · 14 km/h at 15:00` instead of the jarring `Peak 16 km/h at 18:00 · 15:00 · 14 km/h`. The "at" joiner lives in the resource for i18n.
- **Time format.** 24h zero-padded `%02d:00` to match the existing subtitle hours (e.g. `18:00`, `15:00`), replacing the locale-aware `DateTimeFormatter.ofLocalizedTime` which was rendering `3:00 PM` on the same line as the subtitle's `18:00` for en-US users. Indicator always snaps to a whole hour so no minutes are lost. When subtitles gain locale-aware formatting, this should follow.
- **UV unit.** Drop the "UV " prefix from the UV card's readout for parity with its own subtitle (`Peak 7 at 14:00`, not `Peak UV 7 at 14:00`) — the card title already names the metric.

The format lambda passed to `rememberChartReadout` is now `@Composable` so each card can call `stringResource` inside it.

Both scrubbed-state snapshots regenerated and committed.

## Privacy

No change.

## Test plan

- [ ] `./gradlew :app:testDebugUnitTest` — passes locally; two snapshot deltas (`forecast_card_scrubbed.png`, `wind_card_scrubbed.png`) committed.
- [ ] Today screen: subtitle reads `<peak/range> · <value> at <HH:00>` — both halves in the same word order, same time format.
- [ ] UV card readout: just the number, no "UV" prefix.


mikelward:
Scrubbed-state snapshots after the format fix. Subtitle + readout now read in the same word order ("VALUE at TIME"), separated by a mid-dot.

**`app/snapshots/forecast_card_scrubbed.png`**

Before (6de6e116, "TIME · VALUE"):
before

After (8e304af8, "VALUE at TIME"):
after

**`app/snapshots/wind_card_scrubbed.png`**

Before (6de6e116):
before

After (8e304af8):
after

Note: the "before" SHA `6de6e116` is the intermediate commit on the previous PR's branch (since the snapshots were added on #671). The UV-prefix drop isn't visible in these two captures — would need a `UvCardScrubbedPreview` to lock that in; happy to add if you want.

---

mikelward:
After the time-format fix. Subtitle and readout now both render hours in the same 24h zero-padded format.

**`app/snapshots/forecast_card_scrubbed.png`**

Before (8e304af8 — readout in 12h "3:00 PM"):
before

After (adf5684c — readout in 24h "15:00"):
after

**`app/snapshots/wind_card_scrubbed.png`**

Before (8e304af8):
before

After (adf5684c):
after

Also dropped in a `TODO(time-format-locale)` next to `formatScrubHour` covering both the peak-subtitle strings and the readout, with a note about an auto / 12h / 24h override in Settings ▸ Region for users on a 12h locale who prefer 24h (and vice versa) — so both surfaces migrate together when we get to it.

---

PR #693, opened 2026-05-25
Drive the Settings root menu off a single SettingsDest enum

mikelward:
## Summary

The Settings root menu was defined twice — once in `ClothesCastNavHost.settingsMenu(nav)` (runtime, with nav-wired click handlers) and once in `SettingsPreviews.SettingsRootPreview` (with no-op clicks for snapshot rendering). The two drifted: **Format** was added to the runtime menu when the Format settings page landed (#51) and **Developer** earlier, but neither propagated to the preview. Mobile reviewers checking the snapshot couldn't verify those rows.

This PR replaces both call sites with a single ordered `SettingsDest` enum that carries each row's title and subtitle string resources:

```kotlin
internal enum class SettingsDest(@StringRes val titleRes: Int, @StringRes val subtitleRes: Int) {
    SCHEDULE(R.string.settings_root_schedule, R.string.settings_root_schedule_subtitle),
    CLOTHES(...),
    FORMAT(R.string.settings_root_format, R.string.settings_root_format_subtitle),
    // ...
    DEVELOPER(R.string.settings_root_developer, R.string.settings_root_developer_subtitle),
}
```

- **Runtime**: `settingsMenu(nav)` maps `SettingsDest.entries` → `SettingsMenuItem` with each row's nav action. An exhaustive `when` in `NavController.openSettingsDest(dest)` maps each enum value to its destination — adding a new enum value is a compile error until you wire the navigation, which is the whole point of routing every row through the enum.
- **Preview**: `SettingsRootPreview` maps the same `SettingsDest.entries` → `SettingsMenuItem` with no-op clicks. No more hardcoded list to forget to update.

`settings_root.png` regenerates to show all twelve rows (was missing Format and Developer).

## Test plan

- [x] `./gradlew :app:assembleDebug` succeeds
- [x] `./gradlew :app:testDebugUnitTest` passes
- [ ] Inspect the regenerated `settings_root.png` snapshot — should now show Format between Clothes and Location, and Developer at the bottom

---
_Generated by [Claude Code](https://claude.ai/code/session_014bigS97M6bDMxddpNw8K8y)_


mikelward:
## Snapshot diff (1 modified)

`settings_root.png` regenerated to include the previously-missing **Format** and **Developer** rows. Posting inline so it's visible from mobile.

### `app/snapshots/settings_root.png`

Before (`c5f67fe5`, current main):
settings_root before

After (`81ae866f`, this PR):
settings_root after


---

PR #699, opened 2026-05-25
Right-align the scrub readout on chart cards

mikelward:
## Summary

- Replace the `summary · readout` concatenation above every chart card with a two-column row: the static summary (min/max, peak, total) stays in its lane on the left; the live scrub readout sits flush right. The summary stops shifting as the readout appears/disappears, and the eye doesn't have to chase a middle-dot separator to find the current hour's value.
- `appendReadout(subtitle, readout): String?` is gone — replaced by `@Composable ChartSubtitleRow(subtitle, readout)` in `ChartScrubber.kt`, which the four chart cards in `TodayScreen.kt` (feels-like, air, precip probability, precip amount) and the shared `PerModelDiagnosticCard` (wind / cloud / humidity / solar / UV / sunshine) now call.
- The new composable renders nothing when both strings are blank, so dry-day cards without scrub state don't add empty vertical space. Non-scrubbed cards render identically (the row degrades to a single weight-1 Text), so only the two scrubbed previews regenerated.

## Test plan

- [x] `:core:domain:test :core:data:test :app:testDebugUnitTest` — green locally
- [x] Snapshot regen touches only `forecast_card_scrubbed.png` and `wind_card_scrubbed.png` — visual diff confirms summary now stays left while readout sits right
- [ ] CI green
- [ ] Spot-check a long-summary edge case on a narrow screen (the readout's `Spacer(8.dp)` is the only thing keeping them from kissing — weight(1f) on the summary should let it ellipsise/wrap before colliding)

https://claude.ai/code/session_01K9iCyiRq8T1jChmPe7mgPo

---
_Generated by [Claude Code](https://claude.ai/code/session_01K9iCyiRq8T1jChmPe7mgPo)_


mikelward:
## Snapshot diffs

### `forecast_card_scrubbed.png`

Before (`896ec5d`) — readout appended to the summary with a middle-dot:

forecast_card_scrubbed before

After (`5d7dd5c`) — readout right-aligned, summary stays in its lane:

forecast_card_scrubbed after

### `wind_card_scrubbed.png`

Before (`896ec5d`):

wind_card_scrubbed before

After (`5d7dd5c`):

wind_card_scrubbed after

---

PR #738, opened 2026-05-27
Show the correct suburb on the Today card and Location pin

mikelward:
## Summary

The Today card's suburb name and the Location-settings map pin both used to come from coarsening the device fix to a 2dp / ~1 km grid and reverse-geocoding from there. That had two visible problems: the grid-snapped point could land at an arbitrary spot (sometimes outside the recognisable suburb), and the rounding routinely shifted the point across a real suburb boundary so the *name* itself came back wrong.

This PR fixes both at once.

- **Suburb name:** `FetchAndNotifyWorker.resolveLocation()` now reverse-geocodes the un-rounded network-provider fix (~50–300 m, the precision `ACCESS_COARSE_LOCATION` already gives), so the suburb the Today card shows is the one the user is actually in even when they're near a boundary.
- **Map pin:** the reverse-geocoded `addressDetail` ("Brunswick, VIC 3056, Australia") is forward-geocoded through `OpenMeteoGeocodingClient` and snapped to Open-Meteo's canonical centroid for that locality when one's within 10 km of the device fix. The persisted lat/lon is that centroid, so the address-row maps deep link (now the dedicated open-in-maps affordance after #735) lands at the suburb centre rather than a 2dp grid point.
- **No metro-centroid drift:** does *not* broaden to the bare `city` query when the addressDetail lookup misses. `CityNamePicker` intentionally collapses regions to a broader label (NYC's five boroughs all return "New York", Cairo's governorate suffix is stripped), so for a Brooklyn user the addressDetail "Brooklyn, NY, USA" is the right narrow query, and falling back to "New York" would let Manhattan's centroid score within the 10 km radius and silently relocate the persisted location across the East River. The city query now fires only as a last resort when addressDetail is unavailable in the first place (Geocoder returned a locality field but no address lines). [Addresses Copilot review comment.]
- **Fallback:** on any failure path — no `addressDetail`/`city` available, no candidate within the sanity radius, geocoding API error, AOSP-only device with no reverse-geocode backend — we fall back to coarsening the device fix at 2dp, so the persisted-coord precision contract is unchanged on the failure path.
- **Drops downstream rounding now that everything upstream is coarse by construction:**
  - Bug-report `Location: ...` lines stop applying their own `%.2f` to the already-2dp-or-centroid persisted value.
  - The `Locations are rounded to the nearest 1 km for your privacy.` footnote (added in `8b61565`) is removed — it described the prior always-coarsen flow and isn't true any more: the persisted coord is the *suburb centroid* in the happy path, not a 1 km grid snap. PRIVACY.md still spells out the actual contract.

## Privacy

The only material change is that Android's `Geocoder` now receives the raw network-provider fix instead of the 2dp-coarsened one. `ACCESS_COARSE_LOCATION` already caps that precision at city-block level (which is the contract the app declares); the previous 2dp coarsening was an extra belt that hurt name correctness. `PRIVACY.md`'s "Approximate location" section is updated to spell out the city-block-precision Geocoder lookup and where the persisted coord actually comes from.

Nothing else broadens:
- Open-Meteo forecast call — still receives 2dp-or-centroid precision.
- Persisted location — still 2dp-or-centroid; `SettingsRepository.parseLocation()` re-applies `.coarsened()` on read as a belt-and-braces.
- Bug-report coordinate output — now reads the persisted value as-is rather than re-rounding; same precision floor.
- The only *new* outbound payload is the forward-geocode search string ("Brunswick, VIC 3056, Australia") to Open-Meteo — narrow by design, never the broader `city`.

## Test plan

- [x] `:core:domain:test :core:data:test :app:testDebugUnitTest` pass locally.
- [x] `:app:assembleDebug` passes locally.
- [x] `SuburbCentroidResolverTest` covers: nearest-in-radius wins, NYC-borough no-broadening regression, city as last-resort when addressDetail is null, both-queries-null no-op, every-candidate-outside-radius rejection, network-failure no-broadening, all-failure returns null.
- [x] `LocationResolverTest` still passes unchanged.
- [x] `settings_location` Roborazzi snapshot regenerated — the privacy-note footnote is gone; rest of the page (address-row maps link, Refresh button, full privacy policy link) is preserved from #735.
- [ ] On-device: Today card shows the user's actual suburb after a refresh, including near-boundary cases that were misnamed before.
- [ ] On-device: Location settings address row opens maps at the suburb centre.
- [ ] On-device: airplane-mode mid-refresh falls through to the 2dp-coarsened-fix path with no user-visible regression.


mikelward:
Screenshots of the affected surfaces. No snapshot pixel diffs in this PR — both screens already rendered at 2dp / suburb-centroid precision because that's what the preview fixtures use, so the bytes don't change. The user-visible change is at **runtime**:

- **Today card:** the location name in the header (here `Boston`) is now correctly the suburb the user is in, even when the un-corrected 2dp rounding would have shifted across a boundary into a neighbouring suburb.
- **Location settings:** the coords next to the Map button now read the persisted value verbatim (suburb centroid or 2dp coarsening) rather than re-rounding it via `%.2f` on top. The displayed string is identical for any production-realistic input (since the input is already 2dp upstream), but the format no longer lies about doing the rounding.

### Today insight card (`today_insight_card_with_location.png`)

Today insight card with location

### Location settings (`settings_location.png`)

Location settings page

---

mikelward:
Rebased onto current `main` (which landed #735's address-row maps deep link, Refresh button, and the `Locations are rounded to the nearest 1 km for your privacy.` footnote). Dropped the footnote because it's no longer accurate — the persisted coord is the Open-Meteo suburb centroid in the happy path, not a 1 km grid snap. The address-row maps link, Refresh button, and full privacy policy link from #735 are preserved unchanged.

### Location settings · before (footnote still present from #735)

Before: settings_location with privacy footnote

### Location settings · after (footnote dropped)

After: settings_location without privacy footnote

### Location settings · manual · before

Before: settings_location_manual with privacy footnote

### Location settings · manual · after

After: settings_location_manual without privacy footnote

---

mikelward:
Correction — the previous comment's "after" image pointed at SHA `1902a7c`, but that SHA's committed snapshot was the merge-inherited copy from main (still with the footnote) rather than the regenerated one. Force-pushed `1881534` with the actual regenerated snapshots squashed in. The correct "after" renders:

### Location settings · after (`1881534`)

Location settings without privacy footnote

### Location settings · manual · after (`1881534`)

Location settings manual without privacy footnote

---

PR #741, opened 2026-05-27
Tap a manually-set location to open it in maps

mikelward:
## Summary

PR #734 made the reverse-geocoded address line on the Location settings page tappable to open the user's chosen maps app at the 2dp-coarsened coords — but only via the `addressDetail` row, which is populated solely by the device-location path. Manual / forward-geocoded picks leave `addressDetail` null and the line is omitted, so a user who set their suburb via the manual search dialog had **no tap target at all**.

This PR hangs the same maps deep link off the summary line when `addressDetail` isn't there to carry it: render the summary in the primary colour and make it clickable when `current != null && current.addressDetail == null`, so the manual flow gets the same affordance as the GPS flow.

When `addressDetail` is present (device-location path), the summary stays a plain label and the existing `addressDetail` line keeps the action — avoids stacking two visually identical tap targets on top of each other.

Detecting / unset states stay as plain labels — they have no coords to open.

## Snapshot coverage

- `SettingsLocationPreview` (existing) covers the GPS path with `addressDetail` populated — unchanged because the new code path is a no-op when `addressDetail != null`.
- `SettingsLocationManualPreview` (new) covers the manual-pick path — captures the new tappable summary.

## Privacy

Nothing new leaves the device. The maps intent uses the same `~1 km`-coarsened coords (`Location.coarsened()`) the rest of the app already operates on. Same affordance and same coords as PR #734; just a different trigger surface.

## Test plan

- [x] `:core:domain:test` + `:core:data:test` + `:app:testDebugUnitTest` pass locally
- [x] Existing `settings_location` snapshot unchanged (GPS path)
- [x] New `settings_location_manual` snapshot generated and committed
- [ ] CI green

---
_Generated by [Claude Code](https://claude.ai/code/session_01U5EsdG1Hjk1tMLxeSfEe2m)_


mikelward:
Snapshot preview (inlined since the Files tab shows "Binary files not rendered" on mobile):

**Added — `settings_location_manual.png`** (fea78a27)

settings_location_manual

The "Carlton" suburb-name line renders in the primary colour and is the tap target that opens the user's chosen maps app at the coarsened coords — same affordance the GPS path got in #734, now hung off the summary line for the manual-pick case where `addressDetail` is null.

The existing `settings_location.png` (GPS path with `addressDetail` populated) is unchanged — the new code path is a no-op when `addressDetail != null`, so the summary stays a plain label and the `addressDetail` row keeps the action.


---

PR #756, opened 2026-05-27
Tidier Smart Home settings with a master switch for cast

mikelward:
## Summary

- Adds a **"Send forecast to smart display"** master toggle at the top of the Cast section, gating scheduled casts (independently of "Cast now"). New `castEnabled` preference flows through `computeDeliveryGates` so `willCast` is false when the master is off; the picker + per-period toggles disable accordingly.
- Drops descriptive lead-ins and privacy notes that were just restating what the toggles do:
  - Cast: "Send scheduled ClothesCasts…" and "Audio and image are served…"
  - MQTT: "After each scheduled refresh…" and "When on, the rendered…"
- Shortens the topic-prefix hint to `<prefix>/now/text, <prefix>/now/audio, etc`, interpolating the current topic-field value so it updates as the user types.
- Shortens both skip-phone-speech toggles to **"Skip phone speech"** with the description **"Avoid duplicate announcements at home."**
- Strips the dropped/changed strings from all 46 locales so they fall back to the canonical defaults rather than showing stale long-form copy.

## Privacy

No change to what leaves the device — this is text + a UI gate. The master toggle, when off, suppresses one off-device delivery channel (cast publishes), which is a privacy improvement, not a regression.

## Test plan

- [x] `:core:domain:test` — added a new `DeliveryGatesTest` case asserting `willCast=false` when `castEnabled=false`, regardless of per-period toggles. All existing gate tests pass on the same default `castEnabled=true`.
- [x] `:app:compileDebugKotlin` — clean.
- [x] `:app:testDebugUnitTest` — green; `settings_smart_home` Roborazzi snapshot regenerated to reflect the new master toggle row, shortened topic hint with live prefix, and trimmed descriptions.

---
_Generated by [Claude Code](https://claude.ai/code/session_016P6YLmJazi63J8XQDAupUJ)_


mikelward:
### Snapshot diff: `app/snapshots/settings_smart_home.png`

**Before** (`9c78284`):

settings_smart_home before

**After** (`2920d6e`):

settings_smart_home after


---

PR #788, opened 2026-05-29
Get a morning ClothesCast automatically, no setup needed

mikelward:
## What & why

Make the morning / "daily" ClothesCast on by default so a fresh install starts producing casts without the user having to opt in, and remove the onboarding schedule step now that there's nothing to set up. The Today promo card covers schedule discovery instead.

## Changes

- **Daily cast on by default.** `dailyEnabled` now reads on for a fresh install (absent key ⇒ on; an explicit `false` from the schedule page is still respected). Notify + speak were already the default delivery mode (`NOTIFICATION_AND_TTS`), so a new install gets the morning notification and spoken announcement. The notification permission is still requested during onboarding, and the morning alarm is scheduled on first launch from the default-on preference.
- **Onboarding skips the schedule screen.** Both onboarding footer buttons now finish onboarding and land on Today — there's no schedule step to "Continue" into. Removes the now-dead onboarding-landing plumbing: the `fromOnboarding` nav arg on `ScheduleDest`, the schedule-page **Done** button, and the `finishOnboarding` wiring.
- **Promo card covers it now.** The Today "Automatic ClothesCasts" promo shows until the user *acts* on it — dismissing it (X) **or** following its CTA into the schedule page. It no longer keys off the schedule switches (which would hide it immediately now that the morning slot is on out of the box). Both paths persist via `scheduleCardDismissed`.
- **Evening cast stays opt-in.** `tonightEnabled` is unchanged (default off). The one-time grandfather migration still preserves evening casts for installs that had them under the old default-on behaviour; it no longer touches the morning slot, since that reads on by default.

## Privacy

The morning→evening calendar tie-in (`dailyMentionEveningEvents`) now **defaults off** instead of on. It was already gated behind the calendar master switch and `READ_CALENDAR`, so this only *tightens* what can reach the insight prose (and therefore Gemini TTS) — less calendar-derived content leaves the device by default, not more.

## Test plan

`./gradlew :core:domain:test :core:data:test :app:testDebugUnitTest :app:assembleDebug` — all green locally (full CI parity, Android SDK + Google Maven available in sandbox). Updated the repository default/migration tests and the insight tie-in fixture (which now opts the tie-in in explicitly, since it's off by default).

https://claude.ai/code/session_01Y8ZfaxRNpYC1aEnFEuuaRw

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y8ZfaxRNpYC1aEnFEuuaRw)_


mikelward:
### Snapshot regen (`390374e3`) — not from this PR

The bot regenerated **11** `today_insight_card*` snapshots. ⚠️ This is **not** caused by the daily-on / onboarding change — it's a **stale baseline on `main`** being corrected.

The previous `main` commit, `388d5369` "Options to make the ClothesCast shorter", added the `periodPreamble` / `wearPreamble` options (default `ALWAYS` → the *"Today, it will be …"* lede) but landed **without regenerating `today_insight_card.png`**. So `main`'s committed PNGs show the old terse prose while `main`'s code already renders the lede. My PR's snapshot run is the first to record the current rendering, so the delta surfaced here.

All 11 carry the identical change: the prose gains the `Today, it will be` / `Wear a` preambles. Representative before/after:

**`today_insight_card.png`**

| Before (`803b7d38`, stale) | After (`390374e3`) |
|---|---|
| old | new |

**`today_insight_card_dark.png`**

| Before (`803b7d38`) | After (`390374e3`) |
|---|---|
| old | new |

**`today_insight_card_long.png`**

| Before (`803b7d38`) | After (`390374e3`) |
|---|---|
| old | new |

The other 8 share the same preamble delta: `today_insight_card_large_font`, `_layered`, `_location_unknown`, `_with_chevron_both`, `_with_chevron_dark`, `_with_chevron_left`, `_with_chevron_right`, `_with_location`.

Nothing to action on this PR — the regenerated snapshots match current `main` code. Worth a separate cleanup so `main`'s `388d5369` baseline matches its own output, but that's out of scope here.

---

PR #793, opened 2026-05-29
Show notification toggle as off until permission is granted

mikelward:
## What

On the Schedule screen, the **Notify** toggle (in both the morning *Daily ClothesCast* and evening *Nightly ClothesCast* cards) previously reflected the stored delivery mode regardless of whether the OS notification permission was actually granted. So on Android 13+ with `POST_NOTIFICATIONS` denied, the toggle could read **on** while every notification we posted was silently dropped — the toggle lied about what would happen.

This change gates the displayed toggle state on the live permission grant:

- **Permission not granted → the Notify toggle reads off**, regardless of the saved preference (the OS would drop the posts anyway, so "on" was never true).
- **Toggling it on requests the `POST_NOTIFICATIONS` permission.** The row only settles on once the grant lands; if denied it falls back to off. On pre-Android-13 the permission is implicit, so the toggle behaves exactly as before.
- The grant state **refreshes on resume**, so granting/revoking from system Settings is reflected without an in-app action (same approach `NotificationPermissionBanner` already used).
- The recoverable-grant banner now keys off the stored notify *intent* rather than the effective state, so it still appears to guide the user to grant when they want notifications.

## How

`ScheduleContent` now owns a resume-refreshing `notificationGranted` state and threads it through `DayCard` / `NightCard` into `DeliveryModeSection`, where `notifyOn = storedNotify && notificationGranted` drives the switch. The just-in-time permission request on toggle-on (already present) now updates that shared state via the launcher result.

## Snapshots

- `settings_schedule.png` — updated: now captured with the permission **granted** (happy path), so the notification toggles read on and there's no banner.
- `settings_schedule_notifications_blocked.png` — **new**: the same screen with the permission **denied**, showing the Notify toggle off and the "Notifications are blocked / Grant" banner. Backed by a new `SettingsScheduleNotificationsBlockedPreview` so the `no_preview_without_snapshot` guard stays satisfied.

## Privacy

No change to what leaves the device — this is purely UI state derived from a local runtime-permission check.

## Test plan

- `./gradlew :core:domain:test :core:data:test :app:testDebugUnitTest :app:assembleDebug` — green locally (CI parity).
- Roborazzi snapshots regenerated and visually verified for both the granted and blocked states.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WLV937z4G9KTRKXUjAWznE)_


mikelward:
Good catch — fixed in `2c75e51`. The speech (TTS) toggle's `onCheckedChange` now writes the *stored* notify bit (`notifyStored`) rather than the permission-gated display value (`notifyOn`), so touching speech while notifications are permission-blocked no longer drops the saved notification preference or the recovery banner. Granting the permission afterwards re-enables notifications as expected.

---

PR #802, opened 2026-05-29
Chart widgets fill the available space instead of letterboxing

mikelward:
## Problem

On a wide home-screen placement, the feels-like chart widgets didn't use the extra width — the chart sat in the middle with empty gaps on either side. Both `FeelsLikeWidget` and `SevenDayFeelsLikeWidget` rendered their chart to a **fixed 720×560 bitmap** (`SizeMode.Single`) shown with `ContentScale.Fit`, so any cell wider than ~1.29:1 letterboxed the bitmap rather than widening the chart. The chart's height was also pinned at a fixed 180dp, so it couldn't grow/shrink with the cell.

## Change

- **Size the bitmap to the cell.** `buildChartBitmap` reads the widget's actual cell extent (dp) from the `AppWidgetManager` options bundle in `provideGlance` (via `GlanceAppWidgetManager.getAppWidgetId`) and renders at that size, converted to px and bounded to 240–1600px.
- **Clamp the aspect to 2:1 … 4:1.** Between those bounds the bitmap takes the cell's own aspect; outside, it pins to the nearer bound (Glance pads the short side under `Fit`). Keeps the line chart comfortably wide without getting uncomfortably long-and-thin.
- **Let the chart fill its height.** `ForecastChart` / `ForecastCard` gain a `fillHeight` flag (default `false`, so the in-app cards keep their fixed 180dp chart inside the scrolling column). The widget passes `fillHeight = true` so the chart scales with the bitmap instead of wrapping a fixed height.
- Resizing the widget fires an options-changed update that re-runs `provideGlance`, re-rendering at the new size. Falls back to a mid-range 3:1 when no size is reported yet (picker preview).

`SizeMode.Single` is retained — the size comes from the options bundle, not `LocalSize`.

## Snapshots

Adds `feels_like_widget_wide` — a 3:1 fill-height preview locking in the new look. The in-app card snapshots are byte-identical (default `fillHeight = false` path unchanged).

## Privacy

No change to what leaves the device — off-screen render dimensions only.

## Test plan

- `./gradlew :app:compileDebugKotlin` — green.
- `./gradlew :app:testDebugUnitTest` — green (new wide snapshot recorded; existing snapshots unchanged). Note: `SettingsViewModelTest` is a known pre-existing flake under full-suite ordering (`Dispatchers.Main` isolation) — passes in isolation, unrelated to this change.
- The rasterisation path (`ComposeRender.kt`) is on-device only; worth a visual check on a wide cell (chart fills width), a tall/square cell (stays wide, centred), and resize (re-renders).

https://claude.ai/code/session_013U54123Mkv52GUrvDrs3A9


mikelward:
New snapshot added in `e86f2c8` — the wide (3:1) fill-height widget look:

feels_like_widget_wide

The chart now spreads to fill a wide cell instead of letterboxing in a fixed-aspect box, clamped to between 2:1 and 4:1.

---

PR #821, opened 2026-05-30
test: guard every outfit tier has a render-path defaults entry

mikelward:
## What

Adds an `entries`-driven exhaustiveness test so a new `OutfitSuggestion.Top` / `OutfitSuggestion.Bottom` tier can't be added without its corresponding entry in the hand-maintained render-path defaults maps (`outfitTopDefaults` / `outfitBottomDefaults`).

## Why

Every outfit render surface — the Compose home-screen icons (`GarmentTopIcon` / `GarmentBottomIcon`), the rasterised notification + widget bitmap (`renderOutfitBitmap`), and the Nest-Hub card (`renderOutfitCard`) — resolves a garment tier's baked fill/stroke through `outfitTopDefaults.getValue(top)` / `outfitBottomDefaults.getValue(bottom)`. `getValue` throws `NoSuchElementException` on a missing key, so a new tier added without its defaults entry compiles cleanly and only blows up **at render time** — on the home screen, in the widget, and in the cast card simultaneously.

That's the same hand-maintained-parallel-structure gap that bit the gloves work (a new garment feeds one structure but skips another). It's the render-path sibling of:
- **#813** — guarded the phraser's `GARMENT_RES_IDS` map
- **#815** — guarded the snapshot `*DeltaC` fields

The `topDrawable` / `bottomDrawable` resolvers are already compiler-guarded (exhaustive `when`s with no `else`), so the missing-drawable case can't compile; the defaults `Map`s are the unguarded ones, so this drives the check off `entries` to close that gap.

## How

Pure-JVM test (the defaults maps are plain Kotlin — no Robolectric), mirroring `AppPaletteTest`'s `.getValue`-lookup guard style:
- every `OutfitSuggestion.Top` is present in `outfitTopDefaults`, and the map carries no extras;
- every `OutfitSuggestion.Bottom` is present in `outfitBottomDefaults`, likewise.

A non-empty result names the offending tier so the fix (add its `GarmentDefaults`) is obvious. When a future third slot lands (e.g. an `OutfitSuggestion.Hands` gloves icon — `ic_outfit_gloves` already exists), its `outfitHandsDefaults` map gets a sibling assertion here.

## Scope

Test-only (one new file under `src/test/`) — `test:` subject prefix, filtered from the Play "What's new". No production change.

## Test plan

`:app:testDebugUnitTest --tests "*OutfitRenderExhaustivenessTest"` — green (7 tops / 5 bottoms all covered, so the assertions are meaningful). Verified locally against the Android SDK.

https://claude.ai/code/session_01HMjFrx4mng9MUrQ4ZDdKLX

---
_Generated by [Claude Code](https://claude.ai/code/session_01HMjFrx4mng9MUrQ4ZDdKLX)_


mikelward:
The snapshot-regen bot added **2 new baseline PNGs** in `8920e5c` — and they're unrelated to this test-only change. The gloves preview *tests* (`clothing_gloves`, `clothing_thick_jacket_with_gloves`) landed on `main` in commit `7392eee`, but their baseline PNGs were never committed, so the first CI run on any branch regenerates them. That happened to be this branch. They're correct baselines; embedding here since binary diffs don't render in the GitHub mobile app:

**`clothing_gloves.png`** (new)

clothing_gloves

**`clothing_thick_jacket_with_gloves.png`** (new)

clothing_thick_jacket_with_gloves

No action needed on the icons themselves — this just fills a pre-existing baseline gap on `main`. Flagging in case you'd rather these baselines land via a separate `test:` commit instead of riding along here.

---

PR #822, opened 2026-05-30
Point smart-home and cast errors at likely network or firewall causes

mikelward:
## What

Follow-up to the delivery-failure banners (#814). When MQTT or Cast delivery fails, the error message now helps the user tell a **network / firewall** problem from a **service** problem, instead of dumping a raw exception string.

Before, a failed MQTT publish surfaced something like:

> `AnnotatedNoRouteToHostException: No route to host`

Now it surfaces:

> Couldn't reach the broker at `192.168.1.50:1883` — the phone may be on a different network, or a firewall is blocking it. (`NoRouteToHostException: No route to host`)

The raw exception is kept in parentheses for the bug report; the lead is plain English.

### Hints by cause

| Cause | MQTT message leads with… |
|---|---|
| DNS / unknown host | name + "check the address, and that this phone is on the same network as your broker" |
| No route | "the phone may be on a different network, or a firewall is blocking it" |
| Connection refused | "check it's running and listening on port N" |
| Timeout | "the broker may be offline, or a firewall is silently dropping the connection" |
| TLS handshake | "check the \"Use TLS\" setting matches your broker and that its certificate is valid" |

**Cast:** its typed failures already carried network wording (display not found on the network, no LAN address, display didn't fetch — check the LAN). This adds the same hint to the generic *unexpected-failure* fallback (e.g. the phone's media server couldn't bind / be reached). Non-network failures (broker ACL rejection, etc.) are surfaced unchanged.

## How

- New `app.clothescast.net.classifyNetworkError` walks a throwable's cause chain (depth-bounded, cycle-safe) and classifies it as `UNKNOWN_HOST` / `NO_ROUTE` / `CONNECTION_REFUSED` / `TIMEOUT` / `TLS`, or `null` when it isn't a connection-layer failure.
- Matches on exception **type**, not message text, so it's robust across Android versions and locales. HiveMQ/Netty wrap socket errors in `Annotated…` subclasses of the `java.net` types, so the `is` checks catch them; a suffix-based class-name fallback covers any non-subclassing wrapper.
- `MqttPublisher` and `CastInsightController` map the classification to domain-specific wording (broker vs. display) and fall back to the raw message when it returns `null`.

## Privacy

No change to what leaves the device. These strings are built on-device from local socket/TLS exceptions plus the user's own configured broker host/port, and are shown in the same on-device surfaces (Today banner, Smart Home settings, bug report) as before.

## Test plan

- New `NetworkErrorHintTest` covers each kind, the cause-chain walk, the Netty-wrapper name fallback, the refused-vs-no-route split, the non-network → `null` case, and the depth bound.
- New `MqttPublisherTest` cases assert the friendly hint (host:port + firewall guidance) appears for a no-route/unknown-host failure, that a non-network broker rejection keeps its raw message, and that `networkHintFor` wording covers each kind.
- `./gradlew :core:domain:test :core:data:test :app:testDebugUnitTest :app:assembleDebug` — green.

https://claude.ai/code/session_01BCdgkSPetCxDH6xYP5Wddn

---
_Generated by [Claude Code](https://claude.ai/code/session_01BCdgkSPetCxDH6xYP5Wddn)_


mikelward:
The snapshot bot's `ci: regenerate UI snapshots` (`a3509fb`) added two **new** snapshots — `clothing_gloves` and `clothing_thick_jacket_with_gloves`. These are unrelated to this PR's network-hint change: the gloves garment previews already exist in `GarmentDrawablePreviews.kt` on `main`, but their PNGs were never committed, so the first run on any branch that triggers a regen fills the gap. Harmless, `ci:`-prefixed so it stays out of the Play changelog. Inline for phone review (added files, new version only):

**clothing_gloves**

clothing_gloves

**clothing_thick_jacket_with_gloves**

clothing_thick_jacket_with_gloves

---

PR #834, opened 2026-05-30
Add US English rule for strings and commit messages

mikelward:
## What

Adds a new `## Language` section to `AGENTS.md` (symlinked as `CLAUDE.md`) codifying that we author all English in US spelling and idiom.

## Why

The base `values/strings.xml` locale is effectively `en-US` — it's the fallback for the `en-rGB` / `en-rAU` / `en-rZA` translations — and CI funnels every commit subject straight into `whatsnew-en-US`, the US English Play Store "What's new". A stray Britishism (`colour`, `cancelled`, `grey`) in either place ships to users, so the rule calls both out explicitly along with PR text, code comments, and docs. Localized non-English `values-*` files are explicitly exempt.

## Scope

Docs-only change to the agent guide. No code or string changes.

## Test plan

N/A — documentation only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TNuYmYS3cEo6TGWYWdz5kG)_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -329,6 +329,27 @@ new rule the first time something bites you, not the third.
   applicationId `app.clothescast`, download `google-services.json`,
   drop it at `app/google-services.json`, and rebuild. Don't commit it.
 
+## Language
+
+- **Use US English everywhere.** Base English strings
+  (`app/src/main/res/values/strings.xml`), commit subjects and bodies, PR
+  titles / descriptions, code comments, and docs all default to American
+  spelling and idiom: `color` not `colour`, `center` not `centre`,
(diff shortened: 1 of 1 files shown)


mikelward:
Good catch — fixed in `cb509ea`. The original wording wrongly implied `en-rAU` / `en-rZA` fall back to base `en-US`. They actually inherit `en-rGB` (British spelling) via Android's API 24+ locale fallback, as documented in those `strings.xml` headers. The rule now (a) describes base `values/` only as the *final* fallback, and (b) adds an explicit exception: don't "correct" British spelling in any `values-en-*` file, and don't assume an untranslated en-AU/en-ZA string resolves to US English.

---

PR #843, opened 2026-05-30
Add a Play button to the Preview your ClothesCast card

mikelward:
## What

The "Preview your daily ClothesCast" promo card now carries its **own Play button** instead of only pointing the user at the top app bar's. And the trailing "now" is dropped from the title.

| Before | After |
|---|---|
| Title: "Preview your daily ClothesCast **now**", body tells you to find the toolbar Play button, no CTA on the card | Title: "Preview your daily ClothesCast", with a **▶ Play** button right on the card |

## How

- `PlayPromoCard` / `PlayPromoCardContent` gain an `onPlay` callback (+ an `enabled` flag) and render a filled Play button under the body.
- The action is wired in `BannerStack` (which already has `TodayState` in scope) using `LocalContext` + the existing `triggerPlay` — no new callback threaded through the layout layers. It mirrors the top-bar Play button exactly:
  - derives the period (TONIGHT vs TODAY) from the wall clock at tap time,
  - retires the promo after playing, but only once a forecast has been delivered (same `hasForecast` gate that renders the card, so a fresh install can't persist the dismissal early),
  - disabled while `anyWorkActive`.
- `today_play_promo_title` loses the trailing "now".

## Privacy

No change to what leaves the device.

## Test plan

- `./gradlew :core:domain:test :core:data:test :app:testDebugUnitTest` — green; `play_promo_card` snapshot regenerated.
- `./gradlew :app:assembleDebug` — green.

### Snapshot

play promo card

https://claude.ai/code/session_01LhCwbSYoZ47ptq1WjkmyYE

---
_Generated by [Claude Code](https://claude.ai/code/session_01LhCwbSYoZ47ptq1WjkmyYE)_


mikelward:
### Play promo card: before → after

GitHub's mobile app shows "Binary files not rendered" for PNG diffs, so here's the change inline.

| Before (`main`) | After |
|---|---|
| before | after |

Title drops the trailing "now", and the card gains its own ▶ Play button (same action as the top-bar Play).

---

PR #863, opened 2026-05-30
Reorder the outfit and forecast sections on the home screen

mikelward:
## What

Lets the user **drag to reorder** the two home-screen sections — the **Outfit** preview and the **Forecast** insight — from a new **Home screen layout** card on **Settings → Display**. Each row has a grab handle; dragging commits the new order. MVP scope: reordering only, no hide/add — but the design is built to extend.

## Why

First step toward customizable home-screen sections. Scoped to reordering the two sections that already exist.

## How

- **`HomeSection` enum** (`:core:domain`) is the single source of truth for which sections exist, with a `normalize()` helper that de-dupes, preserves the user's order, and appends any missing sections in default order — so a section added to the enum later automatically shows up for users whose stored order predates it.
- Persisted as a normalized JSON list in DataStore (`SettingsRepository.setHomeSectionOrder` / `parseHomeSectionOrder`, mirroring the clothes-rules pattern) and threaded through `UserPreferences` → `TodayState` / `SettingsState`.
- `HomePageScaffold` renders the outfit row and an `insightSlot` in `homeSectionOrder` order, before the always-trailing chart/placeholder `content()`. `TodayPage` passes its insight card + confidence chip as the slot; `SevenDayPage` passes its week-headline card — so the order applies on **all three pager pages** and the outfit row keeps a consistent vertical offset across swipes.
- The default order `[OUTFIT, INSIGHT]` reproduces the previous layout byte-for-byte (only the `settings_display` snapshot changed).
- **Drag-to-reorder** uses the `sh.calvin.reorderable` Compose library (`ReorderableColumn` + a `draggableHandle()` grab handle). Added to the version catalog (`reorderable = "3.1.0"`, Maven Central) and `:app` dependencies. The settled drag commits via `SettingsViewModel.reorderHomeSection(from, to)`.

## Privacy

No change to what leaves the device — purely a local display preference.

## Test plan

- New domain test `HomeSectionTest` covers `normalize()`; new repo tests cover default / round-trip / partial-list normalization.
- `:core:domain:test`, `:app:testDebugUnitTest`, and `:app:assembleDebug` all green locally (with the new dependency resolving).
- `settings_display.png` snapshot regenerated for the new card (drag handles); Today/7-day snapshots unchanged.

## Known limitation

Drag handles aren't reachable via TalkBack drag — a follow-up could add custom accessibility reorder actions. Flagging since it's a regression from a keyboard/AT perspective vs. tappable arrows; happy to add accessible move actions if you'd prefer.


mikelward:
### Snapshot change: `settings_display`

The only snapshot affected is the Settings → Display page, which now carries the new **Home screen layout** card. Today / 7-day snapshots are unchanged (default order is byte-identical to before).

**Before** (`88fab01`):

settings_display before

**After** (`6baeaa8`):

settings_display after

---

mikelward:
### Snapshot updated: `settings_display` (now with drag handles)

Switched the reorder UI from up/down arrows to **drag handles** (supersedes the arrow version in the earlier snapshot comment). Today / 7-day snapshots remain unchanged.

**Before** (`dcefee8` — no card):

settings_display before

**After** (`c18d9de` — Home screen layout card with drag handles):

settings_display after

---

PR #864, opened 2026-05-30
Show each home page's own conditions in the at-a-glance strip

mikelward:
## Problem

The conditions strip showed the **same values on all three home pages** — Today, Tonight, and 7-day all displayed this period's feels-like range (e.g. "13–24°C" three times). It was pinned to `outfitInsight` (always this period), copied from the outfit row's deliberate "same pair on every page" behavior — but for a single-value conditions readout that's just wrong.

## Fix

The strip now reads the hourly series of the page the user is actually looking at, via a new `conditionsHourly` parameter on `HomePageScaffold`:

- **Pages 0 / 1** use that page's own period insight (this period, then the next period) → each shows its own feels-like range, rain, wind, UV.
- **7-day page** summarizes the whole window from its flattened 168-hour stream → the strip reads the week's feels-like range and peak rain / wind / UV.

The outfit row stays pinned to this period's insight as before — it intentionally shows the same today + tonight pair on every page.

7-day strip shows the week range

The 7-day strip now reads "6–22°C · 40%", matching the "Feels like 6–22°C" chart header below it.

## Privacy

No change to what leaves the device — still on-device rendering of the already-fetched forecast.

## Test plan

- `:core:domain:test` and `:app:testDebugUnitTest` — green.
- The 7-day preview's sample no longer needs a hand-tuned outfit-insight hourly series; the strip there now derives from the week's sample days, so the four `seven_day_page*` snapshots are regenerated to show the week range.

https://claude.ai/code/session_01EsckiR3EeJJX3UCayxJdqw

---
_Generated by [Claude Code](https://claude.ai/code/session_01EsckiR3EeJJX3UCayxJdqw)_


mikelward:
### Snapshot diff

The 7-day strip now reads the **week's** range instead of repeating this period's figures. Before is `main` (`62e0388`), after is this branch (`a54e67b`):

**seven_day_page**
| before — `main` (`62e0388`) | after (`a54e67b`) |
|---|---|
| before | after |

The `before` strip showed `9–18°C · 60% · 32 km/h · UV 7` (an artifact of the sample's hand-tuned outfit-insight hourly); `after` shows `6–22°C · 40%`, derived from the week's sample days and matching the "Feels like 6–22°C" chart header below. The other three `seven_day_page*` snapshots changed identically.

Note: the per-period pages (Today / Tonight) aren't snapshotted through the full scaffold — their previews render `InsightCard` directly — so this fix's effect there is verified by the data-source change rather than a pixel diff.

---

PR #865, opened 2026-05-30
Reorder the confidence chip and chart deck on the home screen too

mikelward:
## What

Extends home-screen section reordering (Settings → Display → Home screen layout) to cover **all** the page content. On top of the existing conditions / outfit / insight, two more sections are now reorderable:

- **Confidence** — the forecast-confidence chip, now its own card instead of riding along inside the insight card.
- **Charts** — the whole chart deck (feels-like, air temp, precipitation, and the per-model diagnostics), reordered as one contiguous block.

The two new sections **append after the existing order**, so the default stays **Conditions → Outfit → Forecast → Confidence → Charts** — conditions remains at the top of the page, with the confidence chip sitting directly above the charts.

## Why

Continues the home-customization work. Splitting confidence out of the insight card was a prerequisite you called out; folding charts in completes "reorder all the content."

## How

- `HomeSection` gains `CONFIDENCE` and `CHARTS`; they append to `DEFAULTS` after the existing `CONDITIONS, OUTFIT, INSIGHT`. `normalize()` is unchanged — it just appends any omitted values, so a stored order self-heals (no migration; there are no users yet).
- `HomePageScaffold` gains `confidenceSlot` and `chartsSlot` alongside the existing `insightSlot`, and renders each section in `homeSectionOrder`. Each slot no-ops when it has no data this period (confidence needs a confidence summary; charts need hourly data), so an ordered-but-dataless section simply contributes nothing.
- `TodayPage` supplies all three slots — the confidence chip moved out of the insight slot (keeping its tap-to-reveal-spread + scroll wiring), and the chart deck moved verbatim into `chartsSlot` so it keeps its single `ChartScrubController` / `SpreadCoordinator` / y-axis envelope. `SevenDayPage` supplies its week-header (insight) + chart deck; no confidence chip there.
- The confidence chip's tap now scrolls to whichever of the chip / chart deck sits **higher** in the layout, so when Charts is reordered above Confidence the revealed per-model overlays still land on screen (the deck's top offset is captured via `onGloballyPositioned` on a transparent wrapper column). Default order keeps the previous behavior.
- Settings list grows to five draggable rows (Conditions, Outfit, Forecast, Confidence, Charts).

## Deferred (TODOs in code)

- **Hiding** sections (show/hide), not just reordering.
- Reordering **individual charts** within the deck (kept contiguous so they keep sharing the scrub controller) — the deck is one "Charts" row for now.

## Privacy

No change to what leaves the device — purely a local display preference.

## Test plan

- `HomeSectionTest` updated for the new defaults/append behavior; `SettingsRepositoryTest` round-trip/normalize tests updated for the five-section order.
- `:core:domain:test`, `:app:testDebugUnitTest`, `:app:assembleDebug` all green locally.
- Snapshots regenerated: `settings_display` (five rows) and the `seven_day_page` set; `feels_like_widget_week` picks up the current render (matches CI). Today-page snapshots are per-component and unaffected.

## Note

Drag handles still aren't operable via TalkBack drag (carried over from the earlier reorder PR) — accessible move actions remain a possible follow-up.


mikelward:
### Snapshot changes

Binary diffs don't render on the GitHub mobile app, so here are the affected snapshots inline. The default order changed to **Outfit → Forecast → Conditions → Confidence → Charts**, so the conditions strip moves below the week header on the 7-day page and the Settings list grows to five rows.

**Settings → Display** — before (`62e0388`) → after (`cae885f`):

settings before
settings after

**7-day page** — before (`62e0388`) → after (`cae885f`):

seven-day before
seven-day after

The three `seven_day_page_with_*` variants changed the same way (conditions strip repositioned below the header).

---

mikelward:
Thanks @codex — fixed in `53c3833`. The chip's tap now scrolls to whichever of the chip / chart deck sits **higher** in the layout (`minOf(chipScrollOffset, chartsScrollOffset)`), so when Charts is moved above Confidence the revealed per-model overlays land on screen. The chart deck's top offset is captured via `onGloballyPositioned` on a transparent `Column` wrapper (same `spacedBy`, so no visual change). Default order (chip above charts) keeps the previous behavior since the chip is then the higher of the two.

---

mikelward:
### Rebased onto `main` (`c70dd17f`) — updated snapshots

Rebased onto the latest `main` (which gained "Show each home page's own conditions in the at-a-glance strip", `c70dd17f`). Resolved a conflict in `HomePageScaffold`'s signature (kept `conditionsHourly` **and** the new `confidenceSlot` / `chartsSlot`) and regenerated the seven_day snapshots on the new base. Binary diffs don't render on mobile, so here they are inline — base `c70dd17f` → head `5f19a3f`:

**Settings → Display:**

settings before
settings after

**7-day page:**

seven-day before
seven-day after

The three `seven_day_page_with_*` variants moved the conditions strip below the week header the same way. `:app:testDebugUnitTest` + `:app:assembleDebug` green on the rebased tree.

---

mikelward:
### Updated: conditions strip kept at the top

Reordered the default so the **Conditions** strip stays at the top of the page — the two new sections now append after the existing order: **Conditions → Outfit → Forecast → Confidence → Charts**. Updated `HomeSectionTest` / `SettingsRepositoryTest` for the new default and regenerated snapshots. Binary diffs don't render on mobile, so here they are inline — base `c70dd17f` → head `ac0ef4c`:

**Settings → Display** (Conditions now first):

settings before
settings after

**7-day page** (conditions strip back at the top, above the week header):

seven-day before
seven-day after

`:core:domain:test` + `:app:testDebugUnitTest` + `:app:assembleDebug` green locally.

---

PR #872, opened 2026-05-31
Don't play a voice preview for Gemini until an API key is set

mikelward:
## Problem

On the Speech (Voice) settings page, switching the voice engine to **Gemini** — or tapping a Gemini voice / style / locale — auto-fires a voice preview. With no API key configured, that preview can't reach Gemini, so `runTtsPreview` catches the failure and **falls back to the on-device voice**. The result: the device voice plays as if it were the Gemini voice the user just selected, giving a misleading impression of what Gemini sounds like.

## Fix

`VoiceContent` now suppresses the preview entirely whenever Gemini is selected and no key is configured. A small `canPreview(engine)` guard in the shared `preview()` helper covers every trigger at once:

- the engine radio (switching to Gemini),
- the Gemini voice / style pickers,
- the accent (locale) picker,
- the Test Voice button.

The **Test Voice button is also disabled** in that state, so it reads as "enter a key first" rather than a dead control — the existing "Not set" key status above it explains why. Previews resume automatically once a key is saved.

Unchanged: the device engine path, and the Gemini path once a key is configured.

## Privacy

No change to what leaves the device. This *narrows* outbound traffic slightly — a preview is no longer attempted (and no Gemini request fired) when there's no key. Nothing new is sent.

## Tests

- `./gradlew :app:testDebugUnitTest` — green (933 tests).
- Added a "Settings · Voice (Gemini, no key)" Roborazzi preview + snapshot covering the disabled Test button, and its `no_preview_without_snapshot` meta-test entry.

https://claude.ai/code/session_01MUxDABeJji23M8ezR9pp8r

---
_Generated by [Claude Code](https://claude.ai/code/session_01MUxDABeJji23M8ezR9pp8r)_


mikelward:
New snapshot added in this PR — `settings_voice_gemini_no_key.png` (Gemini selected, no key). The **Save Gemini key** button is disabled (empty field) and the **Test Voice** button further down is disabled too, so no device-voice fallback masquerades as Gemini:

settings_voice_gemini_no_key

---

PR #887, opened 2026-05-31
Show that the Gemini API key is optional in Voice settings

mikelward:
## Summary

At v1301 there was no on-screen signal that the Gemini voice could be used without pasting an API key. Selecting **Settings → Speech → Gemini** led with the API-key field and the status text "Get a free API key at aistudio.google.com", strongly implying a key was required — even though the shared-key proxy (4da91e9e) has made the Test voice button work without one for a few commits now.

When the shared-key proxy is reachable on this build, the Voice settings card now reframes the key field as optional:

> Optional — paste your own key. Get one free at aistudio.google.com

Builds without the proxy (forks / CI without `GEMINI_PROXY_URL` or Firebase) keep today's copy unchanged — there a key really is required.

Mirrored in the **Speech setup** bottom sheet so the first-opt-in flow from Schedule / Smart home matches.

The Test voice button itself was already enabled in this state since 4da91e9e (`enabled = geminiKeyConfigured || sharedTtsAvailable`); this PR doesn't change that wiring, only the copy that explains it. See the full-page snapshot in the PR comment for confirmation that the button shows as enabled.

## Changes

- New string `settings_api_key_status_unset_shared`.
- `VoiceSettings.kt` + `SpeechSetupSheet.kt` pick it when `sharedTtsAvailable && !geminiKeyConfigured`.
- Plumb `sharedTtsAvailable` from `SettingsViewModel` state through `ScheduleSettings` / `SmartHomeSettings` into `SpeechSetupSheet`; `VoiceSettings` already had it.
- New `SettingsVoiceGeminiNoKeySharedPreview` + matching Roborazzi entry so the new copy is snapshotted at the standard 640dp viewport.
- New `SettingsVoiceGeminiNoKeySharedFullSnapshot` test class (same pattern as `SettingsClothesFullSnapshot`) capturing the same preview at h2400dp so the whole Voice engine card fits in one frame — the standard viewport crops just below the voice picker, hiding the Test voice button which is the very thing the new copy unlocks.
- Translation of `settings_api_key_status_unset_shared` into the 44 `values-*` overrides tracked as a follow-up in `docs/TODO.md`.
- Layout reorder (voice / Test voice up top, API key collapsed below) also tracked as a follow-up — out of scope for this PR.

## Test plan

- [x] `./gradlew :app:testDebugUnitTest` passes locally (Android SDK at `/opt/android-sdk`).
- [ ] CI JVM-unit-tests + assembleDebug green on push.
- [ ] Snapshot `settings_voice_gemini_no_key_shared_full.png` shows the new "Optional — paste your own key…" status and the **enabled** (filled blue) Test voice button at the bottom of the card.
- [ ] Manual: with the shared proxy reachable, open Settings → Speech, pick Gemini, confirm new copy. Tap Test voice — plays without pasting a key.
- [ ] Manual: with the proxy disabled (clear `GEMINI_PROXY_URL`), confirm the old "Get a free API key…" copy still appears and Test voice stays disabled until a key is entered.

https://claude.ai/code/session_01GvJxixiyb9cF8zEgsx8ood


mikelward:
Snapshot added (`4f61ff5`):

**`settings_voice_gemini_no_key_shared.png`** — new

settings_voice_gemini_no_key_shared

Captures the new "Optional — paste your own key. Get one free at aistudio.google.com" status text on builds where the shared-key Gemini proxy is reachable. Companion to the existing `settings_voice_gemini_no_key.png`, which still shows the old "Get a free API key at aistudio.google.com" copy for builds without the proxy.

---

mikelward:
Re-shot at full page (`457e908`) so the whole Voice card fits in one frame — the standard 640dp viewport cropped just below the voice picker, hiding the Test voice button which is exactly the thing the new copy is supposed to unlock.

**`settings_voice_gemini_no_key_shared_full.png`** — new (added in this push)

settings_voice_gemini_no_key_shared_full

You can see the **Test voice** button at the bottom is enabled (filled blue, not greyed) — same enable rule as before (`geminiKeyConfigured || sharedTtsAvailable`) but now with the "Optional — paste your own key…" copy above to tell the user why they can tap it without entering anything.

Also filed a follow-up TODO ([`docs/TODO.md`](https://github.com/mikelward/clothescast/blob/457e908da7f6cb05aeb7e6ae4ed58b412c8510a7/docs/TODO.md)) to reorder this section: voice / style / Test voice up top, API key field collapsed below as the "unlimited use" upgrade. As the snapshot shows, the key entry still dominates the prime real estate just under the engine subhead, even though the user can already hear Gemini for free by hitting Test voice — leading with "try it now" instead of "paste a key" is the right shape, just out of scope for this PR.

---

PR #935, opened 2026-06-01
Upgrade to AGP 9.2.1 / Gradle 9.4.1, and theme chart axis labels

mikelward:
Two `internal:` commits, infra first. Verified with the full CI-equivalent build (`:core:*:test`, `:app:testDebugUnitTest`, `:app:assembleDebug`) on JDK 21.

## 1. `internal:` Upgrade to AGP 9.2.1 / Gradle 9.4.1
| | From | To |
|---|---|---|
| AGP | 8.10.1 | 9.2.1 |
| Gradle wrapper | 8.11.1 | 9.4.1 |
| Roborazzi | 1.32.0 | 1.63.0 |

Migration changes AGP 9 / Gradle 9 forced:
- **Built-in Kotlin.** AGP 9 ships Kotlin support, so the standalone `kotlin-android` plugin is removed from `:app` and the root build (and its catalog alias dropped). Compiler options move from `kotlinOptions {}` inside `android {}` to a top-level `kotlin { compilerOptions { } }`. The `:core:*` modules keep `kotlin.jvm`.
- **Roborazzi ≥ 1.57 required** — AGP 9 removed the `TestedExtension` roborazzi 1.32 applied against.
- **JUnit Platform launcher** — Gradle 9 no longer injects it, so `junit-platform-launcher` is now `testRuntimeOnly` in `:app`, `:core:domain`, `:core:data`.

**Snapshot harness hardening (`PreviewSnapshots.kt`).** Vico runs its `CartesianChartModelProducer` transaction on `Dispatchers.Default`, which the pinned Compose clock / `waitForIdle()` don't wait on. Under the parallel test load the newer roborazzi timing exposed, the chart model could land *after* capture, rasterizing the plot blank — and since the snapshot job is record-only, that would pass CI silently. The capture loop now yields real wall-clock time between captures while a chart is still settling (chart-less previews with a baseline still return on their second capture without sleeping), bars an under-sized "chrome only" capture from passing the stability check, and keeps the existing baseline rather than committing a blank over a good chart if one never settles in budget.

Snapshots: roborazzi 1.63 captures dialog previews with their scrim/backdrop instead of tightly cropped → 4 dialog/picker PNGs regenerated. All chart snapshots stayed byte-identical to `main` under the fixed harness.

## 2. `internal:` Theme chart axis labels with onSurface
The forecast charts' hour/value ticks used Vico's default axis label color — a fixed ~54%-opacity black that isn't derived from the Material color scheme. **Readable and unchanged on the light card** (no user-visible difference on the promoted light builds), but low-contrast on the dark card. Now colored with the theme's `onSurface` via a shared `rememberOnSurfaceAxisLabel()` helper on every forecast chart's start + bottom axes. Regenerates 39 chart snapshots; the only visible change is higher-contrast tick labels in dark mode.

## Test plan
- [x] `./gradlew :core:domain:test :core:data:test :app:testDebugUnitTest` — green
- [x] `./gradlew :app:assembleDebug` — green
- [x] Charts render (not blank) under CI-equivalent parallel load
- [x] Axis labels: light unchanged, dark higher-contrast (verified visually)
- [ ] CI parity


mikelward:
### Snapshot changes (4 dialog/picker previews)

Roborazzi 1.63 now captures dialog previews **with their scrim/backdrop** instead of tightly cropped — same content, fuller framing. All chart snapshots are byte-identical to `main`. Before = `main` (`b5bdf774`), after = this PR (`1c1c0e01`).

**bug_report_consent_dialog**
| before | after |
|---|---|
| before | after |

**outfit_rationale_dialog**
| before | after |
|---|---|
| before | after |

**outfit_rationale_dialog_tuned**
| before | after |
|---|---|
| before | after |

**settings_garment_color_picker**
| before | after |
|---|---|
| before | after |

---

PR #961, opened 2026-06-02
Hide chance-of-rain legend chips for models with no rain-chance line

mikelward:
## What

The "Chance of rain" card listed a per-model legend chip for **every** model in `PerModelHourly.byModel`, but the chart only plots models that have a non-null `precipitation_probability`. The result: a model with no probability data (UKMO, JMA, …) showed a **legend chip with no line** — the "where's the UKMO line?" report.

This fix makes the chance-of-rain legend mirror the chart's data filter, exactly as the sibling **rainfall-amount** card already does.

## Root cause (not Vico)

The recent Vico 2.5.1 → 3.2.1 bump was a red herring. Vico 3.2.1 renders the tail model fine once it has data — verified by adding UKMO to the snapshot fixture and confirming the line draws.

The real issue is data shape + a legend/chart filter mismatch:

- **UKMO** comes from the side-band multi-model call, which reads `precipitation_probability_ukmo_seamless` — a field Open-Meteo **doesn't populate** for UKMO. So UKMO lands in `byModel` (it has temperature + precipitation amount) but its probability series is wholesale null → the chart correctly skips it, but the legend didn't.
- **Auto** (`best_match`) is unaffected: its series comes from the primary `/v1/forecast` call, which **does** carry the top-level `precipitation_probability` (Open-Meteo's GFS-ensemble-derived product, served regardless of the auto-picked deterministic model). So Auto keeps both its line and its chip.

`TodayScreen.kt` had three `byModel`-based legend filters; only the chance-of-rain one was wrong. The two temperature cards are correct as-is (every model in `byModel` has a temperature line), and the diagnostic cards already share one filtered list between chart and legend.

## Test fixture

Added a UKMO entry to `SAMPLE_PER_MODEL_HOURLY` modeling the real Open-Meteo shape — temperature + `precipitation_mm` present, probability `null`. This exercises the tail of `MODEL_DRAW_ORDER` (the ECMWF/GFS/ICON trio never reached it) so this class of legend/chart mismatch can't regress unnoticed. After the change, UKMO draws on the temperature and rainfall-amount cards (and their legends) but is correctly absent from the chance-of-rain card.

## Snapshots

- `*_with_model_spread` cards gain the UKMO line (temp/amount) or correctly omit it (chance-of-rain).
- `*_consensus` cards shift slightly because the cross-model mean now includes UKMO in the sample.

## Privacy

No change to what crosses the device boundary — UI legend filtering and preview sample data only.

## Test plan

- `./gradlew :core:domain:test :core:data:test :app:testDebugUnitTest` — green; snapshots regenerated.
- Verified `precipitation_card_with_model_spread.png` (no UKMO chip/line) and `forecast_card_with_model_spread.png` (UKMO line + chip present).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Geh2qfJLVZrxUo65uKtmb4)_


mikelward:
### Snapshot diffs (GitHub's Files tab won't render binary PNGs / mobile)

The clearest way to read this change is the contrast between two cards in the **new** fixture (which now includes a UKMO model with temperature + rain-amount but *null* rain-probability — the real Open-Meteo shape):

**Chance of rain** — UKMO correctly **absent** from line and legend (legend now mirrors the chart filter; this is the fix):

`before d46d1df` → `after 879fddc`

chance-of-rain before
chance-of-rain after

**Feels-like temperature** — UKMO line + chip now **present** (it has temperature data, and this proves Vico 3.2.1 renders the tail model fine):

`before d46d1df` → `after 879fddc`

temperature before
temperature after

The other 20 changed PNGs are the same fixture effect: `*_with_model_spread` cards gain a UKMO line where the metric exists, and `*_consensus` cards shift slightly because the cross-model mean now includes UKMO.

---

PR #974, opened 2026-06-03
Show the umbrella's name under the outfit, not just its icon

mikelward:
## What

Fixes the reported bug: on the Today screen the outfit **image shows an umbrella, but the caption text under it doesn't name it** — the caption no longer matches the image on the same card.

## Root cause

The Today screen shows the two outfit cards **side by side at half width**. The caption was a single `Text` capped at `minLines = 2, maxLines = 2`, with the worn outfit forced onto line 1 and accessories onto line 2:

```
Sweater · Rain jacket · Jeans
Umbrella
```

When the worn line is long enough to wrap on a narrow card (e.g. a rainy default setup that lights a rain jacket), it soft-wraps to two lines and consumes the whole budget — so the hard `maxLines = 2` **clips the `Umbrella` line**. The figure holds an umbrella the caption never names.

## Fix

Drop the forced worn/accessory break. The caption is now **one flowing `· `-joined run** of every piece, which simply soft-wraps, so the trailing accessory is never clipped:

```
Sweater · Rain jacket · Jeans · Umbrella
```

- **Today card:** no line cap — it grows. `minLines = 2` keeps a short caption from being cramped, and both cards in the row are stretched to equal height (`Modifier.height(IntrinsicSize.Max)` + `fillMaxHeight()`) so a wrapping card doesn't run taller than its neighbour.
- **Widget:** can't grow into a fixed cell, so it reserves/caps the right number of lines via a shared `outfitGarmentCaptionLineCount` (2 once any extra piece joins the base top+bottom, else 1).
- **One source:** both surfaces now build the text from a single `outfitGarmentCaption` helper — the old line-splitting variant is gone, so the card and widget can't drift apart.

Display-only — nothing here crosses the device boundary.

## Test plan

- New `outfit_row_rain_jacket_umbrella` snapshot reproduces the report's case (rain jacket + umbrella) at half width, locking the regression.
- The accessory outfit-card snapshots (`outfit_coat_pants_gloves`, `outfit_coat_pants_gloves_umbrella`, `outfit_jacket_jeans_umbrella`) and the matching widget snapshots (`widget_today_coat_gloves_umbrella`, `widget_today_coat_pants_gloves`, `widget_today_jacket_umbrella`) re-record to the flowing caption. Accessory-less captions are byte-identical, so all other snapshots are untouched.
- `./gradlew :app:testDebugUnitTest` and `:app:assembleDebug` pass locally.

Image diffs posted as a follow-up comment (binary diffs don't render on mobile).

https://claude.ai/code/session_0194M5L8m2RhBcngrMPtUXMk


mikelward:
New snapshot added by this PR (binary diffs don't render on mobile, so inlining it). Left "Today" card has the wrapping worn line + umbrella — the `Umbrella` line that the old `maxLines = 2` cap used to clip is now shown:

outfit_row_rain_jacket_umbrella

---

mikelward:
Snapshot image diffs for the flowing-caption fix (mobile hides binary diffs, so inlining). Before = `main`, after = `2858f8d4`.

**The reported case — `outfit_row_rain_jacket_umbrella`** (new, half-width side-by-side). Left "Today" card holds the umbrella with a wrapping worn line; the umbrella the old `maxLines = 2` cap clipped is now named and wraps naturally:

row after

**`outfit_coat_pants_gloves_umbrella`** — full-width card, caption now one flowing run instead of a forced two-line split:

| Before (`main`) | After (`2858f8d4`) |
|---|---|
| before | after |

**`outfit_coat_pants_gloves`** and **`outfit_jacket_jeans_umbrella`** changed the same way (forced break → flowing run):

| `outfit_coat_pants_gloves` after | `outfit_jacket_jeans_umbrella` after |
|---|---|
| cpg | jju |

---

mikelward:
Unified the card and widget onto one `outfitGarmentCaption` helper — the widget now renders the same flowing caption. Updated snapshots (after = `a20ba859`):

**Widget** — `widget_today_coat_gloves_umbrella`, now flowing like the card (`Thick coat · Long pants ·` / `Gloves · Umbrella`):

widget

Other re-recorded widget snapshots:

| `widget_today_coat_pants_gloves` | `widget_today_jacket_umbrella` |
|---|---|
| cpg | ju |

The Today-card snapshots (`outfit_row_rain_jacket_umbrella`, `outfit_coat_pants_gloves*`, `outfit_jacket_jeans_umbrella`) are unchanged from the previous push — same flowing caption, now sharing the same helper as the widget.

---

PR #968, opened 2026-06-03
List every garment and accessory under the outfit icons

mikelward:
## What

The caption under the outfit icons listed only the top and bottom (e.g. "Sweater · Jeans"), so the rain-jacket outer layer, gloves, and the carried umbrella went unnamed even though the icon clearly shows them. This makes the caption name **every** piece the figure wears or carries, on **both the Today card and the home-screen widget**, with the widget styled to match the card.

## Behavior

The worn outfit (top → rain jacket → bottom) reads on the first line; accessories (gloves, umbrella) drop to a second line:

| Outfit | Caption |
|--------|---------|
| Sweater + jeans | `Sweater · Jeans` (one line) |
| Sweater + jeans + rain jacket | `Sweater · Rain jacket · Jeans` (one line) |
| Coat + pants + gloves + umbrella | `Thick coat · Long pants` / `Gloves · Umbrella` (two lines) |

Shared via a small `outfitGarmentCaption` helper (reuses the existing localized `garment_*` strings).

## Widget matches the card

- **Label:** the widget's "Today" now uses the same muted `onSurfaceVariant` grey as the card's period label (was full-strength `onSurface`).
- **Caption spacing:** kept as a single two-line Text so the lines sit tightly together (a single TextView spaces its own lines by the font line spacing; two stacked Texts each add padding and read loose). The Compose preview mock pins a snug line height to match.
- **Card** already reserves two caption lines (`minLines = 2`); the **widget** reserves the second line only when an accessory needs it, so plain outfits keep full-size icons.

## Test plan

- Regenerated the affected Today-card and widget preview snapshots.
- `:app:testDebugUnitTest` and `:app:assembleDebug` pass locally.
- Rebased onto `main` (which re-fixed the rain-jacket icon); the rain-jacket card snapshot re-recorded against the new art.

## Snapshots

Image diffs posted as a follow-up comment (binary diffs don't render on mobile).

https://claude.ai/code/session_01PesRoBea2EgJiFXWzWawrG

PR #976, opened 2026-06-03
Keep the umbrella visible in narrow side-by-side widgets

mikelward:
## What

Fixes the accessory-clipping regression in the widget caption (follow-up to #968). When the widget splits into two side-by-side outfits, each column is ~half the widget width, so the flowing garment caption can wrap to a **third** line — e.g. `Thick coat · Long pants · Gloves · Umbrella` in a ~120dp column. The caption was capped at a constant 2 lines (`outfitGarmentCaptionLineCount` = 2 for any accessory outfit), so the final accessory was clipped/ellipsized even though the icon shows it.

## Fix

`outfitGarmentCaptionLineCount` now **derives the cap from the actual wrapped-text budget** instead of a constant: it estimates how many lines the run takes at the column's width and font size (greedily packing the space-separated tokens — counting at word boundaries so a long final word like "Umbrella" correctly tips onto a new line), and the widget reserves icon room and caps `maxLines` to that, up to 3.

- A **single-column** cell keeps the established 1-line base / 2-line accessory wrapping → **existing widget snapshots are unchanged** (the estimate is calibrated to match).
- A **halved side-by-side column** grows to 3 lines so the last accessory stays visible.
- The Today card is untouched (it has no cap and grows freely).

The estimate is a deliberately rough proxy for real glyph widths, biased to not *under*-count (an over-count only adds a little icon-shrinking headroom; an under-count clips). It only frames the text; the renderer still does the actual wrapping.

## Test plan

- New `OutfitCaptionLineCountTest` (pure JVM) pins the calibration: single-column accessory = 2 lines, halved column = 3, long base = 1, plain = 1, capped at 3.
- New `widget_side_by_side_accessories_narrow` preview/snapshot covering a ~120dp accessory column — the umbrella renders, not clipped.
- No existing snapshots changed; `:app:testDebugUnitTest` and `:app:assembleDebug` pass locally.

## Snapshot

Posting the new narrow side-by-side image as a follow-up comment.

https://claude.ai/code/session_01PesRoBea2EgJiFXWzWawrG

---
_Generated by [Claude Code](https://claude.ai/code/session_01PesRoBea2EgJiFXWzWawrG)_


mikelward:
### Narrow side-by-side accessory column (`6f8aa3ff`)

A ~120dp column (240dp widget split in two) now renders the full accessory caption — the umbrella is no longer clipped. Left column is the accessory outfit; right is a plain one.

narrow side by side

---

PR #1008, opened 2026-06-09
Bug hunt: nine fixes across consensus math, casting, settings, and charts

mikelward:
## Summary

A systematic bug hunt across all three modules (four parallel review passes: `:core:domain`, `:core:data`, app services, Compose UI), with the confirmed findings fixed. Nine commits, each scoped to one theme.

### Forecast correctness

- **Synthetic-zero best_match votes** (`OpenMeteoClient`): best_match's per-model series was built from the mapped `HourlyForecast` list, whose null hours the mapper zero-fills for the chart. At the 14-day horizon edge those fake 0 °C / 0 % values were averaged into the consensus blend as real votes. The series is now built from the wire payload with the same null policy as `MultiModelConfidenceFetcher.parseHourly` (drop the hour on null temp, keep precip null). Side effect: best_match now carries the primary call's wind/UV, so it votes in those consensuses and draws on the diagnostic charts (the old omission rested on a stale comment claiming the primary call doesn't fetch them — it does).
- **Rainfall/sunshine consensus undercounting** (`PerModelConsensus`): averaging per-model totals is arithmetically identical to counting a partial model's missing hours as zero — the exact bias the KDoc (in three files) claimed the design avoided. Switched to hour-wise mean over reporting models, then sum; the one test pinning the biased value was updated (3.0 mm → 3.5 mm for the partial-model scenario). **This changes the displayed "X mm of rain / Xh of sun" numbers when a model has a partial series** — flagging since it's a deliberate behavior change to match the documented intent.
- **Model-pruning substring match** (`MultiModelConfidenceFetcher`): a 400 naming `icon_d2_15min` would also prune a requested `icon_d2`. Now whole-id word-boundary matching.

### Charts

- **Per-model lines vs. range band misalignment**: commit 3816359 made the *band* hour-keyed but left every per-model *line* (and the diagnostic cards' consensus mean, the rainfall main line, and the "Peak X at H" subtitles) position-keyed — so exactly when a model drops an hour, lines drew one hour left of the band. Everything now resolves x positions by timestamp lookup against the chart's own hourly window, which also fixes the band drifting an hour off the main line on DST-transition weeks. Rendered output is unchanged in the common case; `ui.today` snapshot tests pass unmodified.

### Casting & scheduled delivery

- `ensureSession` leaked a `SessionManagerListener` on every session start (removed only on cancellation) — one per scheduled cast on the process-lifetime `SessionManager`. Now unregistered on every terminal callback, plus a re-check that closes a registration race causing spurious "did not respond in time" timeouts.
- `castCurrentInsight` converted `CancellationException` into a persisted "cast failed" Settings status when the user left mid-cast; now rethrown.
- `CastMediaServer.publish`/`stop` race (worker thread vs. main-thread `onSessionEnded`) could clear the token a fresh URL was just minted for; now synchronized.
- The notification and phone-speaker `launch`es in `deliver()`'s `supervisorScope` were unguarded — a child failure there reaches the default handler and kills the process. Both now follow the scope's documented best-effort-per-destination policy. A few `runCatching` sites that swallowed `CancellationException` now rethrow.

### Settings robustness & TTS

- One stored clothes rule with an unknown condition type (future-version write + downgrade) reset the **entire** rule list to defaults; unknown types now degrade per-item like unknown garments.
- The two schedule-time strings were the only unguarded parses in `toUserPreferences`; a corrupt value made the preferences flow throw on every emission (app-wide, unrecoverable). Now defaulted.
- `isRetryableGeminiTtsFailure` treated missing/invalid BYOK keys as transient, burning all synth retries with backoff before falling back to the device voice; both now fall back immediately.

### Smaller fixes

- **Activity leak + broken locale switch on Android 12/12L**: ViewModel factory lambdas captured the Activity; after rotation the VM held the destroyed instance and `recreate()` on it was a no-op. Factories now capture only the Application; the locale switch resolves the live Activity via a lifecycle-tracked weak reference.
- **Outfit icon ignored a firing `pants` rule** (no PANTS tier in `pickOutfit`) — prose said pants, icon showed the default bottom. Tier added, rationale cites the pants rule.
- **Inner sandbox build was broken under Gradle 9.4** (the synthetic `:core` project's directory must exist and can't be shared) — fixed with a dedicated `core/parent-project/` directory. The breakage was masked by piping gradlew output through `tail`.
- Four KDoc passages that stated the inverse of (intended, test-pinned) behavior corrected — doc-only.

## Findings reported but not fixed (judgment calls / need verification)

- `readModelDaily` hard-drops a model from the *daily confidence* when `precipitation_probability_max` is missing — asymmetric with the hourly path and possibly starving `gem_seamless` (in the default set) of a confidence vote. Needs verification against the live Open-Meteo API before softening.
- `ensureSession` stuck-route case: a MediaRouter-selected route with no connected session times out on every attempt (nothing re-initiates the session start). Needs a device to validate a fix safely.
- `CastMediaServer` port allocation is a TOCTOU (probe socket closed before Ktor binds), and the server speaks no Range/HEAD — playback works on the Default Media Receiver but seek/preflight degrade.
- 7-day diagnostic "Peak X at 2pm" subtitles don't say which day (the scrub readout does) — polish gap, matches existing "weekly subtitle is a follow-up" notes.

## Privacy

No change to what crosses the device boundary. The consensus-math changes alter locally computed numbers only; rendered insight prose composition is unchanged in structure.

## Test plan

- New regression tests: synthetic-zero blend votes + wind/UV passthrough (`OpenMeteoClientTest`), whole-id pruning (`MultiModelConfidenceFetcherTest`), non-retryable key failures (`GeminiTtsRetryClassifierTest`), hour-wise consensus (`RainfallConsensusTest`/`SunshineConsensusTest`), pants tier + rationale (`OutfitSuggestionTest`), unknown rule type (`SettingsRepositoryTest`), midnight wrap + DST window keying (`RangeBandTest`).
- `:core:domain:test` + `:core:data:test` green (outer and inner builds); `app.clothescast.ui.today.*` (333 tests incl. Roborazzi snapshots) green with snapshots unmodified. Full `:app:testDebugUnitTest` + `:app:assembleDebug` running locally; CI is the authoritative pass.

https://claude.ai/code/session_018MD6LyP3Am3FcED1NCBLw8

---
_Generated by [Claude Code](https://claude.ai/code/session_018MD6LyP3Am3FcED1NCBLw8)_

### DIFF
--- a/app/src/main/kotlin/app/clothescast/ClothesCastApplication.kt
+++ b/app/src/main/kotlin/app/clothescast/ClothesCastApplication.kt
@@ -1,7 +1,9 @@
 package app.clothescast
 
+import android.app.Activity
 import android.app.Application
 import android.content.Context
+import android.os.Bundle
 import app.clothescast.alarm.DailyAlarmScheduler
 import app.clothescast.calendar.CalendarContractEventReader
 import app.clothescast.cast.CastInsightController
@@ -51,6 +53,7 @@ import io.ktor.serialization.kotlinx.json.json
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
+import java.lang.ref.WeakReference
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
 import kotlinx.serialization.json.Json
@@ -266,6 +269,18 @@ class ClothesCastApplication : Application() {
     // Telemetry) and read by the nav layer for the onboarding-skip write.
     internal val applicationScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
 
+    // Weak handle on the currently-resumed Activity, maintained by the
+    // lifecycle callbacks registered in [onCreate]. Lets long-lived lambdas
+    // (e.g. ViewModel factory closures, which outlive any one Activity across
+    // config changes) act on the *live* Activity at call time — the locale
+    // switch's recreate() on API 31/32 — without retaining a destroyed
+    // instance. WeakReference so this never extends an Activity's lifetime.
+    @Volatile
+    private var resumedActivity: WeakReference<Activity>? = null
+
+    /** The Activity currently in the resumed state, or null when none is. */
+    fun currentResumedActivity(): Activity? = resumedActivity?.get()
+
     override fun attachBaseContext(base: Context) {
         // Re-apply the persisted per-app locale before the framework caches a
         // Resources reference for the Application context (used by the worker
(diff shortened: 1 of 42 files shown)


mikelward:
The Codex P1 is a false positive: the brace at `OpenMeteoClient.kt:220` closes the inner `url { ... }` block, not the `httpClient.get { }` builder — the `parameter(...)` calls that follow are still inside the request builder. `:core:data` compiles and its tests pass locally (full `:core:domain:test` + `:core:data:test` + `:app:testDebugUnitTest` + `:app:assembleDebug` run, exit 0); the in-flight CI run on this PR will confirm the same.

---

PR #1021, opened 2026-06-11
Make the update banner buttons easier to see

mikelward:
## What

The "Update available" card on the Today screen draws on a `primaryContainer` background, but its **Update / Restart** filled button and the **Not now / Cancel** text button both default to the `primary` palette — the same family the card is tinted from. Under Material You dynamic color those tones can land close together, so the filled button blends into the card and the dismiss label washes out. That's the "buttons can be hard to see" report.

## Change

Anchor both buttons to `onPrimaryContainer`, the role Material 3 guarantees contrasts with a `primaryContainer` surface:

- **Filled action button** → `containerColor = onPrimaryContainer`, `contentColor = primaryContainer`, so it renders as a solid high-contrast block in every scheme. (Disabled colors were already anchored to `onPrimaryContainer`; this extends the same reasoning to the enabled state.)
- **Dismiss text button** → `contentColor = onPrimaryContainer`, so the label stays clearly legible.

No behavior change — colors only. Regenerated the three update-banner snapshots (`available`, `downloading`, `downloaded`) to match.

## Test plan

- `./gradlew :core:domain:test :core:data:test :app:testDebugUnitTest :app:assembleDebug` — all green locally.
- Snapshot diffs for the three update-banner states are in this PR; image diffs posted as a comment below.

https://claude.ai/code/session_01FvaeEdoMuq92eHpZ3wQj63

---
_Generated by [Claude Code](https://claude.ai/code/session_01FvaeEdoMuq92eHpZ3wQj63)_


mikelward:
Snapshot diffs (GitHub's file view shows "Binary files not rendered", so embedding inline):

**Update available** — before → after
before
after

**Downloading** — before → after
before
after

**Ready to install** — before → after
before
after

The fixed RAINBOW preview palette already had a blue `primary` that contrasted, so the visible shift here is to a darker, higher-contrast pair. The real win is in Material You dynamic-color schemes where `primary` and `primaryContainer` can sit close together — `onPrimaryContainer` is guaranteed to contrast there.

---

PR #1025, opened 2026-06-13
Stop the 7-day rain cards from saying "today"

mikelward:
## What

Follow-up to #1024. On the 7-day chart deck, the **Chance of rain** and **Rainfall** cards still fell back to per-period copy that says "today":

| Card / state | Was (7-day deck) | Now |
|---|---|---|
| Chance of rain — dry | No rain expected **today** | No rain expected |
| Rainfall — dry | No rainfall expected **today** | No rainfall expected |
| Rainfall — total (trace-week fall-through) | 12 mm of rain **today** | 12 mm of rain |

The deck spans a week — and the same composable also renders the **Following 7 days** (days 8–14) page — so "today" was simply wrong there.

The per-period **Today / Tonight** pages are unchanged.

## How

- New `*_week` string variants that drop the time qualifier, selected on the deck via a small `isWeekView()` helper keyed off `LocalScrubMomentFormat` (the same `DayPlusHour` signal the peak subtitles use).
- The qualifier is **dropped** rather than replaced with "this week", because the same deck renders both the next-7 and following-7 windows — "this week" would be wrong on the latter.
- Variants added in the base locale and all ~49 translated locales, derived from each locale's existing today/tonight strings by removing the "today" word.

## Test plan

- `./gradlew :app:testDebugUnitTest` — green; one snapshot re-recorded (`following_week_chart_deck`, Rainfall card now reads "No rainfall expected").
- `./gradlew :app:processDebugResources` / `:app:compileDebugKotlin` — clean.

## Noted, not in scope

The **Sunshine** card's "X of sun today" total has the identical "today" pattern on the deck. Left for a separate change to keep this one focused on the rain cards (which is what was flagged); happy to follow up.

https://claude.ai/code/session_01U1TE5FCDyaU86ro2Z1zWFM

---
_Generated by [Claude Code](https://claude.ai/code/session_01U1TE5FCDyaU86ro2Z1zWFM)_


mikelward:
### Snapshot change (inline, since GitHub mobile can't render binary diffs)

**`following_week_chart_deck`** — the Rainfall card's dry subtitle drops "today" (this sample week is dry). Same change applies to the Chance-of-rain dry state and the Rainfall trace-week total, which this sample doesn't happen to exercise.

Before (`eb7c31e4`):
before

After (`2ac0f4f8`):
after

---

PR #1028, opened 2026-06-13
Rounder, more app-consistent drop-down and overflow menus

mikelward:
## What

Give the open menu overlays a more modern, more app-consistent Material look:

- **Rounder corners** — 12dp menu corners instead of the Material default 4dp.
- **Item text matches the app** — Material's default menu item uses `labelLarge` (the button/label role: medium weight, tight letter-spacing), which reads as a *different font* from the rest of the app. Switched to `bodyLarge` (16sp, normal weight) so menu options match the app's body text. This is the "more like the rest of the app" fix.

Covers the **format / clothes-garment / holiday-override** pickers **and the shared bug-report overflow menu** (⋮), via a single `AppMenuShape` constant. The picker trigger value also bumps to `bodyLarge` so it matches its row label.

## Screenshots / coverage

Open menus render in a `Popup` the screen snapshots can't capture, so this adds two before/after preview mocks (`overflow_menu_style`, `settings_menu_style`) that rasterise the menu container + items inline — both for visible coverage and so the change is reviewable. The closed-trigger change is visible in the updated `settings_format` / `settings_calendar` snapshots.

## Test plan

- `./gradlew :app:testDebugUnitTest` — green (incl. the `no_preview_without_snapshot` guard, updated for the new previews).
- `:app:compileDebugKotlin` — clean.

## Note

`bodyLarge` leans toward "match the rest of the app" over textbook Material (which specs `labelLarge` for menu items) — chosen deliberately per discussion. Corner radius (12dp) and text role are easy to tune.

https://claude.ai/code/session_01U1TE5FCDyaU86ro2Z1zWFM


mikelward:
### Snapshot change (inline, since GitHub mobile can't render binary diffs)

**`settings_format`** — drop-down trigger values now use the larger `bodyLarge` text, matching the row labels. (The rounder 12dp corners and larger menu-item text apply to the open menu, which is a popup and isn't captured in snapshots.) The other three changed snapshots — `settings_calendar`, `settings_format_clothes_never`, `settings_format_current_forecast` — are the same trigger-text change.

Before (`fad27baf`):
before

After (`44feb130`):
after

---

mikelward:
### New snapshot coverage for the open menus (inline, since GitHub mobile can't render binary diffs)

Open menus are popups the screen snapshots can't see, so these mock the container + items to make the change visible. Left/top = **Before** (4dp corners, `labelLarge` button-weight items); right/bottom = **After** (12dp corners, `bodyLarge` body-weight items).

**Overflow menu** (`overflow_menu_style`):
overflow

**Settings drop-down** (`settings_menu_style`):
settings menu

---

PR #1029, opened 2026-06-13
Keep the dev launcher badge visible after adaptive masking

mikelward:
## Why

The local/debug build's launcher icon carries a yellow **DEV** badge so a dev build is easy to tell apart from release at a glance. But the badge was pinned to the bottom edge of the 108dp adaptive-icon canvas — which sits entirely inside the outer 18dp *bleed* the launcher always clips. On any masked launcher, and a Pixel circle especially, the badge was cropped away and the dev icon looked identical to release.

The snapshot preview hid this: it scaled the whole 108dp icon down to fit the 72dp masked viewport (`ContentScale.FillBounds`), which drags bottom-edge content up into the visible circle. A real launcher center-crops the inner 72dp instead, so the scaled preview made a bleed-pinned badge look like it survived when it didn't.

## What

**Commit 1 — faithful preview (`internal:`)**
Render the adaptive-icon layers the way a launcher does: draw each layer at its full 108dp and let a 72dp circular clip *center-crop* the 18dp bleed, rather than scaling to fit. The regenerated `launcher_icon_dev` snapshot now shows the truth — the bottom-pinned badge is cropped away and the dev icon is indistinguishable from release.

**Commit 2 — the fix (`internal:`)**
- Redraw the badge on the full 108dp canvas with its banner and "DEV" text inside the central **safe zone** (the 66dp circle every mask preserves), so it survives a circle / squircle crop. The construction foreground layer-lists now overlay the badge full-size instead of bottom-pinning it.
- Give the dev variant a **monochrome (themed-icon) layer** it previously skipped. It knocks "DEV" out of the lower torso as negative space — the letters are holes in the *same* `evenOdd` path as the shirt, so under themed icons they punch through to the background plate (not the shirt's own tint), the same way the sun is knocked out. With the badge carrying through, the themed dev icon now reads as a dev build, so the old reason to skip monochrome no longer applies.

## Privacy / shipping scope

Dev-only. CI and release builds use the unbadged `@mipmap/ic_launcher` (the badge is gated on `CI != true` in `app/build.gradle.kts`), so nothing here reaches users. No change to anything that crosses the device boundary. Both commits are `internal:`-prefixed and won't surface in the Play "What's new".

## Test plan

- `./gradlew :app:testDebugUnitTest --tests "*PreviewSnapshots.launcher_icon*"` — regenerates the four launcher snapshots; `no_preview_without_snapshot` passes with the new dev-monochrome preview wired in.
- Snapshot images (before → after) posted as a PR comment below, since GitHub's mobile app doesn't render binary diffs in the Files tab.

https://claude.ai/code/session_01HvmUC5FZWSiggRoMYy6gQn

---
_Generated by [Claude Code](https://claude.ai/code/session_01HvmUC5FZWSiggRoMYy6gQn)_


mikelward:
### Snapshot diffs (inline, since binary diffs don't render on mobile)

**Dev icon under a Pixel circle crop — `launcher_icon_dev.png`**

Before (badge pinned to the bottom edge → cropped away, identical to release):

dev before

After (badge in the safe zone → survives the crop):

dev after

**New: dev themed icon — `launcher_icon_dev_monochrome.png`** ("DEV" knocked through to the background plate):

dev monochrome

The release `launcher_icon` and `launcher_icon_monochrome` also re-rendered in the first commit, but only because the preview now center-crops instead of scaling — their artwork is in the safe zone, so the meaning is unchanged.

---

PR #1048, opened 2026-06-20
Drop two upgrade-path migrations that have run their course

mikelward:
## Summary

Drop two upgrade-path migrations that have been live for >2 weeks across all internal-testing builds. Every existing install has long since run them, so the keys / channel are gone from their devices and these calls are now permanent no-ops on the cold start hot path.

## What's removed

- **`weather_alerts_v1` notification channel delete** — added 2026-06-01 (`4b9f9aa8` "Remove the non-working weather alerts feature"). ~20 days live. The retired severe-weather-alerts channel has been removed from every install that ever registered it.
- **`SettingsRepository.clearLegacyHomePreferences()`** + its `Application.onCreate` caller — added 2026-05-26 (`ae26ceab` "Drop the GPS-based at-home TTS gate"). ~25 days live. Wiped the five DataStore keys (`home_latitude`/`longitude`/`display_name`/`country_code`, `skip_tts_at_home`) from upgraders. Skip-phone-speech on MQTT/cast success covers the same UX.

## What's kept

Other migrations are too recent and stay:
- `InsightCache.legacyKeyFor` — 11 days. Forecast-cache survival across the rename.
- `LEGACY_ECMWF_IFS04_NAME` in `SettingsRepository` — 13 days. Pre-0.25° ECMWF model-id forward-migration.
- All channel-ID deletes and the `tonight_insight_fetch` queue cancel from #1046 — landed today.

## Privacy

No change to what crosses the device boundary. Pure on-device cleanup removal.

## Test plan

- [x] `:core:domain:test :core:data:test :app:testDebugUnitTest` green locally
- [x] `:app:assembleDebug` green locally
- [ ] Manual: cold launch on a build that previously had the GPS at-home config — confirm no crash and that there's nothing on disk now to clean (the previous build already did it)

---
_Generated by [Claude Code](https://claude.ai/code/session_01GDuTd2sTYEYTrntVLtBz6t)_

### DIFF
--- a/app/src/main/kotlin/app/clothescast/ClothesCastApplication.kt
+++ b/app/src/main/kotlin/app/clothescast/ClothesCastApplication.kt
@@ -330,15 +330,6 @@ class ClothesCastApplication : Application() {
         // builds, see [castContext] above).
         castInsightController?.bind()
         applicationScope.launch {
-            // One-shot wipe of the home-pin coordinates that the now-removed
-            // GPS-based at-home gate used to persist. Runs every launch but
-            // is idempotent — removes by key name and no-ops when absent —
-            // so the cost on fresh installs / repeat runs is one DataStore
-            // edit's worth of disk I/O.
-            runCatching { settingsRepository.clearLegacyHomePreferences() }
-                .onFailure { DiagLog.w(TAG, "Legacy home-pref cleanup failed", it) }
-        }
-        applicationScope.launch {
             try {
                 val prefs = settingsRepository.preferences.first()
                 // Reconcile Locale.setDefault (process-scoped, lost on cold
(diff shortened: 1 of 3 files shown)


mikelward:
Codex's point is technically correct — a user who skips from a pre-`ae26ceab` build directly to this commit, or who has the cleanup build installed but never launched it, doesn't get the keys removed. Considered and accepting the risk for this PR:

- This is a pre-publishing app with internal FAD testers (`docs/TODO.md` "Pre-publishing blockers" still has unchecked items). Testers auto-update and launch frequently — that's why they're testers. The "skipped-version upgrade" path doesn't really exist here.
- The keys are tiny inert DataStore key-value pairs (a few doubles, strings, one bool). They're not surfaced anywhere user-visible and they're not transmitted off-device — they just sit in the on-device DataStore file as dead bytes.
- The cleanup has been live across every internal build for ~25 days. Any tester who's launched the app in that window has had the keys removed.

For a production Play Store app with millions of irregular-update users this would be worth keeping. For pre-publishing on FAD testers, it's the right cleanup. If we ever do a privacy-sensitive cleanup that needs to be bulletproof across upgrade paths, the right pattern is a one-shot migration tied to a stored schema version, not the indefinite-no-op pattern this was.

---

PR #1053, opened 2026-06-20
Tap a holiday or birthday to see which calendar it's from

mikelward:
## What

In the **Calendar** settings page, the **Public holidays** and **Birthdays** listings now **expand each row on tap** to reveal the **source calendar** the event was read from.

## Why

When a day themes from a detected calendar celebration, there was no way in the app to see *which* calendar produced it — making a surprising entry (e.g. a "King's Birthday" public holiday from a calendar you don't remember subscribing to) impossible to trace. Tapping the row now shows the source account, so you can find and silence it (turn off that calendar in Google Calendar, unsubscribe, etc.).

## How

- `UpcomingCalendarEvent` gains `ownerAccount` (defaulted, so existing call sites are unaffected). The reader populates it from the Calendars-table join it already performs for classification.
- `CalendarCelebrationsSection` rows become tap-to-expand (chevron affordance); expanding shows `Calendar: <account>`.
- New string `settings_holidays_calendar_source`.

## Privacy

The owner account is account/email-shaped PII. It stays **device-local** — shown only on this settings screen, never added to insight prose, TTS payloads, or Firebase, consistent with every other calendar field (`UpcomingCalendarEvent` / `CalendarEvent` doc updated to say so).

## Test plan

- `:app:compileDebugKotlin` + `CalendarContractEventReaderTest` pass; added an assertion that `ownerAccount` flows through `upcomingCelebrations`.
- UI snapshots will be regenerated by CI (rows gain a chevron).
- Manual: Settings → Calendar → expand a Public holidays / Birthdays row → source calendar shown.

## Follow-up

Per-calendar enable/disable setting (defaulting to Google Calendar visibility) is coming as a separate PR — that's the fuller fix for controlling which calendars feed the app.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01HvXeNpUi6S5ibXkdEaeG8g)_


mikelward:
**UI snapshot change** — one file changed: `settings_calendar_celebrations.png`. The celebration rows now carry a tap-to-expand chevron (the source-calendar line only appears in the expanded state, which the static preview doesn't capture).

**Before** (`8dff1a90`):
before

**After** (`a58bfb25`):
after

---

